PR #25, opened 2015-06-11
adding thank you page

ScottDowne:


ScottDowne:
r?

public/thank-you.html
```diff
@@ -0,0 +1,130 @@
...
+              </div>
+              <input name="submit-btn" value="Sign up now" type="submit">
+            </form>
+          </div>
+        </div>
+
+        <script src="https://sendto.mozilla.org/page/-/donation_form/js/jquery-2.1.1.min.js"></script>
+        <script src="https://sendto.mozilla.org/page/-/donation_form/js/parsley.min.js"></script>
+        <script>
+          var
```
alicoding: We should abstract this block into its own file, but next PR would be fine.

PR #29, opened 2015-06-13
an attempted react ssr

ScottDowne:


ScottDowne:
I don't think this is good enough to merge, but it's kinda interesting?

components/link.jsx
```diff
@@ -0,0 +1,13 @@
+var React = require('react');
+var Router = require('react-router');
+var Link = Router.Link;
+
+var Linker = React.createClass({
```
alicoding: Do we need this now then?

ScottDowne: Yeah, I hate remembering how to get to Link every time I need it, which is often, usually requiring me to open up another .jsx file that uses Link and copying and pasting.

Also, I mean to modify this to swap between <a href=""> and <Link to=""> depending on if the link is internal or external. Something like this: https://github.com/mozilla/teach.webmaker.org/blob/develop/components/icon-link.jsx#L6-L22

ScottDowne:
R? This is ready to land, imo.

I would rather land it than keep working on it.

I know there are two demo pages in there that we don't need, but I'll just convert them to sequential and thank you page paths when those pages are made into react.

alicoding:
We will improve this later, but lets land and iterate through this.

PR #30, opened 2015-06-13
handlebars ssr, I would rather compile though

ScottDowne:


alicoding:
I think this method won't work well using handlebars plus it doesn't do that client side comparison like React does with Virtual Dom, so I don't think we should go this path.

ScottDowne:
Able to close this now, we made this as a comparison to building react, react won out and we're now invested in it.

PR #32, opened 2015-06-15
adding simple paypal page and component, right now locked to USD but …

ScottDowne:
…I'll update that later


ScottDowne:
a-rrrrrr plus?

PR #33, opened 2015-06-16
Add static build for react

alicoding:
@ScottDowne maybe you can take a look at this first to see what you think

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,17 +1,19 @@
+
 {
   "name": "donate.mozilla.org",
   "version": "0.0.1",
   "description": "Mozilla's Donation Forms",
   "main": "server.js",
   "scripts": {
-    "start": "npm-run-all --parallel server watch:js watch:css",
-    "build:dev": "npm run build:css",
-    "build:production": "npm run build:css",
-    "build:js": "webpack --progress --profile --colors",
+    "start": "node scripts/build",
+    "start:prod": "node server.js",
+    "start:dev": "npm-run-all --parallel watch:js server webpackserver watch:css",
+    "postinstall": "npm-run-all build:js build:css",
+    "build:js": "webpack",
     "build:css": "npm run watch:css -- --no-watch",
-    "postinstall": "node scripts/build",
-    "watch:js": "npm run build:js -- --watch",
+    "watch:js": "webpack -w -h",
     "watch:css": "mkdirp public/css && autoless less public/css --autoprefix '> 1%, last 2 versions'  --source-map",
+    "webpackserver": "webpack-dev-server --content-base public",
     "server": "node server.js"
   },
   "repository": {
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -48,5 +51,22 @@
     "react-router": "^0.13.3",
     "stripe": "~2.2.0",
     "webpack": "^1.9.10"
+  },
+  "devDependencies": {
```
alicoding: there are bunch of them here I can remove, but will look at them later.

alicoding:
basically right now when you run `npm run dev` it will run our API server and webpack-dev-server for local development and the benefit with that is you get hot loader for js, css and html.

For production we should just build the static file and run our API server which then will generate `main.js` file for client-side rendering if anything ever changed in the state and so on for react to handle, but we will have a nice precompiled static html file for our site \o/

client.jsx
```diff
@@ -1,8 +1,23 @@
...
+var React = require('react')
+var AmpRouter = require('ampersand-router')
+var HomePage = require('./pages/index.jsx');
+var Sequential = require('./pages/sequential.jsx');
+var ThankYou = require('./pages/thank-you.jsx');
 
-var routes = require("./routes.jsx");
+var Router = AmpRouter.extend({
+  routes: {
+    '': 'sequential',
```
ScottDowne: Does this work here too? '/' if not that's fine as is.

alicoding: I can try that

webpack.config.js
```diff
@@ -1,20 +1,40 @@
...
-    filename: '[name].js',
-    chunkFilename: '[id].chunk.js',
-    path: path.join('public', 'build'),
-    publicPath: '/build/'
+    filename: "main.js",
+    path: path.join(__dirname, 'public')
   },
+  hostname: '0.0.0.0',
+  port: 3000,
+  isDev: process.env.NODE_ENV !== 'production',
```
ScottDowne: "NPM_CONFIG_PRODUCTION" Is the production env key we use, because it is default in some places.

webpack.config.js
```diff
@@ -1,20 +1,40 @@
...
+  html: function (data) {
+    var homePageHtmlString = React.createFactory(HomePage)
```
ScottDowne: We should probably find a better name for this, because homepage is the sequential page, and index.jsx is kinda not a correct name either.

PageWrapper maybe?

package.json
```diff
@@ -4,14 +4,15 @@
   "description": "Mozilla's Donation Forms",
   "main": "server.js",
   "scripts": {
```
alicoding: @jbuck can you look at changes in this file to see if you're okay with it?

routes.jsx
```diff
@@ -3,10 +3,10 @@ var Route = require('react-router').Route;
 
 var routes = (
   <Route>
-    <Route name="thank-you" path="/thank-you" handler={require('./pages/thank-you.jsx')} />
-    <Route name="sequential" path="/" handler={require('./pages/sequential.jsx')} />
-    <Route name="share" path="/share" handler={require('./pages/share.jsx')} />
-    <Route name="paypal-donate-usd" path="/paypal-donate-usd" handler={require('./pages/paypal-donate-usd.jsx')} />
+    <Route name="thank-you" path="/thank-you/?" handler={require('./pages/thank-you.jsx')} />
+    <Route name="sequential" path="/?" handler={require('./pages/sequential.jsx')} />
```
ScottDowne: Curious, does '/' need the '?'?

Seems like that makes no sense and is kinda unexpected.

alicoding: Might not, but I don't know how our server handles that, so I just put it anyway...

PR #38, opened 2015-06-16
link up sequential form with thank you page and a few bugs fixed and …

ScottDowne:
…a copy change


ScottDowne:
R?

This just links the thank you page, and adds back some stuff I regressed in the thank you page itself.

PR #49, opened 2015-06-19
Localization

alicoding:


pages/thank-you.jsx
```diff
@@ -37,9 +39,9 @@ var ThankYou = React.createClass({
           <div>
             <div id="header-copy">
               <div className="row">
-                <h1>From all of us at Mozilla</h1>
+                <h1><L10nMSG msgid="ty.From_all_of_us_at_Mozilla"/></h1>
```
ScottDowne: You mentioned there was an html formatted API call here, just a reminder to use it.

pages/sequential.jsx
```diff
@@ -541,7 +543,7 @@ var Sequential = React.createClass({
                 <fieldset id="page-1" className="sequence-page">
                   <div className="row">
                     <div className="full">
-                      <h2>Donate now</h2>
+                      <h2>{this.getIntlMessage("sequential.donate_now")}</h2>
```
ScottDowne: Update this to be similar to the thank-you api usage.

package.json
```diff
@@ -41,7 +42,9 @@
     "json-loader": "0.5.1",
     "mkdirp": "^0.5.1",
     "npm-run-all": "^1.2.5",
+    "object-assign": "^3.0.0",
```
ScottDowne: You can remove this.

locales/index.jsx
```diff
@@ -0,0 +1,26 @@
...
+if (global.window) {
+	locales = getLocales(require.context('./', true, /\.json$/).keys());
+} else {
+	var reqF = require('enhanced-require');
+	var req = reqF(module);
+	locales = getLocales(req.context('./', true, /\.json$/).keys());
+}
+
+module.exports = {
+	locales: getLocales(locales),
```
ScottDowne: Do you need to call getLocales on locales twice?

components/client.jsx
```diff
@@ -1,9 +1,21 @@
...
+import {locales} from '../locales';
+
 
 Router.run(routes, Router.HistoryLocation, function (Handler, state) {
-  React.render(<Handler {...currencies[state.params.currency]}/>, document.querySelector("#my-app"));
+	var values = {};
+	if(state.params.currencies) {
+		values = currencies[state.params.currency];
+	} else {
+		values = {};
```
ScottDowne: values already defaults to {} so you don't need the else

pages/thank-you.jsx
ScottDowne: Oh, I noticed you already changed this in the l10nMSG call.

components/localized-msg.jsx
```diff
@@ -0,0 +1,15 @@
+var React = require('react/addons');
```
ScottDowne: We are not using localized-msg.jsx anymore are we?

alicoding: Nope

alicoding: Ideally we should but I can't figure out out why I need to require mixins
twice yet

components/client.jsx
```diff
@@ -1,9 +1,20 @@
...
 import Router from 'react-router';
 import { currencies } from '../scripts/currencies.js';
-
 import { routes } from './routes.jsx';
+import {locales} from '../locales';
+
 
 Router.run(routes, Router.HistoryLocation, function (Handler, state) {
-  React.render(<Handler {...currencies[state.params.currency]}/>, document.querySelector("#my-app"));
+	var values = {};
```
ScottDowne: Still some weird tabs in this patch.

ScottDowne:
Failing travis.

ScottDowne:
R+ with those changes and a passing travis.

alicoding:
Travis is failing for not having `Intl` library on server side, but I wouldn't worry about this for now. I've merged this since we are going to use nodejs and if we've decided to use iojs our server can enable that.

alicoding:
I'm also looking at polyfill/how to enable this for travis only, but don't want to block it on that.

PR #52, opened 2015-06-20
de wire paths a bit

ScottDowne:


ScottDowne:
paths.js now just handles returning an array of all paths. Including locale paths.

routes.jsx now no longer needs to care about paths. Which isn't needed client side.

scripts/paths.js
```diff
@@ -1,22 +1,26 @@
...
-    name: "give-bitcoin",
-    path: "/give-bitcoin/?",
-    handler: require('../pages/give-bitcoin.jsx')
-  }
-};
+import currencies from './currencies.js';
+import locales from "../locales";
+import pages from "./pages.js";
+
+function getLocalePaths(arr) {
```
alicoding: Lets optimize this one a bit.

 ``` js
function getLocalePaths(paths, locales) {
  var returnArray = [];
  paths.map(function(path) {
    returnArray = returnArray.concat(locales.map(function(locale) {
      return path + locale;
    }));
  });
  return returnArray;
}
```

This way we can just call it like this:

``` js
getLocalePaths(['/sequential-', '/thank-you-'], ['en', 'th'])
// return Array [ "/sequential-en", "/sequential-th", "/thank-you-en", "/thank-you-th" ]
```

You can also take this further by merging `locales` array with `currencies` array and just call this function once like this

``` js
getPaths((['/sequential-', '/thank-you-', '/paypal-donate-'], ['en', 'th', 'USD', 'CAD']);
```

I don't know what you think about this approach?

ScottDowne: Yeah, I like where you went with this.

I still kept them as separate function calls though. I wasn't sure where you were going with
```
getPaths((['/sequential-', '/thank-you-', '/paypal-donate-'], ['en', 'th', 'USD', 'CAD']);
```

We wouldn't want to end up with paths like '/sequential-cad'

PR #55, opened 2015-06-25
some locale work and react work for simple paypal page

ScottDowne:


components/simple-paypal.jsx
```diff
@@ -1,44 +1,68 @@
...
+        {message}
+      </button>
+    );
+  }
+});
 
 var simplePaypal = React.createClass({
+  mixins: [IntlMixin],
+  getInitialState: function() {
+    var value = "";
```
alicoding: I don't think you need to do this?

components/simple-paypal.jsx
```diff
@@ -1,44 +1,68 @@
...
-      updateDonateButtonText(amountSelected);
-    }
-
     // check if a pre-selected amount is passed via the URL
     if (window.location.hash.match(/#amount-\d+?/)) {
-      var preSelected = window.location.hash.substring(8);
-      $theForm.find('[name="donation_amount_other"]').val(preSelected);
-      updateDonateButtonText(preSelected);
+      this.setState({
+        value: window.location.hash.substring(8)
```
alicoding: Preferably to use what react router expose here but its fine for now

pages/paypal-donate.jsx
```diff
@@ -2,8 +2,10 @@ var React = require('react');
 var Footer = require('../components/footer.jsx');
```
alicoding: Do you mind changing all these to be es6 so its consistent?

alicoding:
R+

components/simple-paypal.jsx
ScottDowne: I tried and wasn't super happy with it.

It requires more work.

```
console.log(Router.HashLocation.getCurrentPath())
```
vs
```
console.log(window.location.hash);
```

Not only that, but `Router.HashLocation.getCurrentPath` doesn't work server side, it fails on getting `window.location.hash`...

Feels super redundant to me.

PR #86, opened 2015-07-21
Tests

ScottDowne:


package.json
```diff
@@ -13,7 +13,9 @@
     "build:css": "npm run watch:css -- --no-watch",
     "watch:js": "webpack -w -h",
     "watch:css": "mkdirp public/css && autoless less public/css --autoprefix '> 1%, last 2 versions'  --source-map",
-    "server": "node server.js"
+    "server": "node server.js",
+    "test": "npm-run-all --parallel start test:selenium",
```
alicoding: I think we will need a way to close the server somehow too?

alicoding:
Your test is failing and I left one comment there.

package.json
ScottDowne: Yeah, agreed, but I'm not sure how yet.

ScottDowne:
Odd, why is Travis running the tests? I didn't tell it to do that yet and no idea if it would event work.

alicoding:
r+ but file to get this run by travis.

PR #91, opened 2015-07-27
Optimizing L10N

alicoding:
Changes to this patch and why:

1. Change from JSON to YAML
  This is just an improvement in translation flow for our translators.
Using YAML will allow us to add comments for our translators in case the
strings is unclear.

2. Added bulk require for preloading all the translation files
  This will help reduce error and runtime issue for loading our
translation files.

3. Remove namespace for pages.
  We used to namespace all of our strings, but since we are not a big
application I don't think that's useful, yet.


webpack.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
 
   module: {
     loaders: [
+      { test: /\.js$/, loaders:  ['babel-loader'], exclude: ['node_modules'] },
```
ScottDowne: Can you explain the diff between loading babel for js$ and jsx$?

package.json
```diff
@@ -48,5 +49,13 @@
...
     "stripe": "~2.2.0",
     "webpack": "^1.9.10"
+  },
+  "devDependencies": {
+    "colors": "^1.1.2",
+    "fs-readfile-promise": "^2.0.1",
+    "fs-writefile-promise": "^1.0.1",
+    "js-yaml": "^3.3.1",
+    "xml2json": "^0.7.1",
+    "yaml-intl-xml-json-converter": "0.0.7"
```
ScottDowne: What are each of these doing?

It's not clear where in the code these are loaded in, I wonder if some of these are from testing?

PR #102, opened 2015-07-28
Fix l10n issue

### LINKED ISSUE #99: Localized pages not working, or fully localized
See example for DE:

screenshot 2015-07-28 14 19 42

* Input fields missing
* Flow between steps broken
* Most text not localized

alicoding:
fixes #99

In Router component the expected object is
``` js
{

  locales: [...],
  messages: {...}
}
```

but I mistakenly passed only the messages object without its parents, so
this is why it failed to load the page.

I'm going to document all these l10n related stuff in another ticket and
make sure it's in the README.


alicoding:
@ScottDowne R?

PR #108, opened 2015-07-28
Add documentation for localization

### LINKED ISSUE #56: Add Documentation for Localization into README
Can we please include a brief section in readme.md outlining the steps we would need to complete to add a new locale.

alicoding:
fixes #56

This is my first write up on localization. Hopefully I didn't miss
anything here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,3 +16,60 @@ $> cp sample.env .env
 ```
 $> npm start
 ```
+
+## Localization
+
+In this project we're using [React-Intl](https://github.com/yahoo/react-intl) to localize our application and YAML for translation.
+
+#### Localize a component or page
+
+To localize a component or page you have to always include IntlMixin in your class mixins for example:
+
+``` typescript
+var React = require('react');
+var IntlMixin = require('react-intl').IntlMixin;
+
+var Example = React.createClass({
+	mixins: [IntlMixin]
+  render: function() {
+    return (
+      <div>
+      	<h1>{this.getIntlMessage('key_name_here')}
+      </div>
+    );
+  }
+
+});
+```
+
+Once you added the mixin it will expose `getIntlMessage` method for your component to get the localized message for the given key name.
+
+#### Adding locale
+
+Since we are using YAML for our translation, but React-Intl expect JSON which will require an extra build step to convert from YAML to JSON.
+We are using [yaml-intl-xml-json-converter](https://www.npmjs.com/package/yaml-intl-xml-json-converter) to convert from YAML to JSON.
+
+##### config for for YAML to JSON conversion
+
+`intl-config.json`
+``` json
+{
+	"supportedLocales": ["en-US", "de", "fr", "pt-BR", "es"],
+	"dest": "locales",
+	"src": "locales",
+	"type": "json"
+}
+```
+
+##### YAML template
+
+`en-US.yaml`
+``` yaml
+---
+en-US:
+  first: This is your first message
+  second: This is your second message
+```
+
+You have to make sure you match your language code in your YAML file and the name of the file with what you include in your config file for the converting part otherwise it will fail.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,3 +16,60 @@ $> cp sample.env .env
...
+
+`en-US.yaml`
+``` yaml
+---
+en-US:
+  first: This is your first message
+  second: This is your second message
+```
+
+You have to make sure you match your language code in your YAML file and the name of the file with what you include in your config file for the converting part otherwise it will fail.
```
ScottDowne: I think we can simplify this line, it's a bit wordy and hard to read.

Something like:

"The language code and title of the YAML file must match a locale in the `supportedLocales` intl-config.json array"

That is, if I understand what you were initially trying to say.

README.md
```diff
@@ -16,3 +16,60 @@ $> cp sample.env .env
...
+  }
+
+});
+```
+
+Once you added the mixin it will expose `getIntlMessage` method for your component to get the localized message for the given key name.
+
+#### Adding locale
+
+Since we are using YAML for our translation, but React-Intl expect JSON which will require an extra build step to convert from YAML to JSON.
```
ScottDowne: Maybe better?

"Because we are using YAML for our translation and React-Intl expects JSON, we need an extra build step to convert YAML to JSON"

README.md
```diff
@@ -16,3 +16,60 @@ $> cp sample.env .env
...
+#### Localize a component or page
+
+To localize a component or page you have to always include IntlMixin in your class mixins for example:
+
+``` typescript
+var React = require('react');
+var IntlMixin = require('react-intl').IntlMixin;
+
+var Example = React.createClass({
+	mixins: [IntlMixin]
```
ScottDowne: Some white space we can remove.

README.md
```diff
@@ -16,3 +16,60 @@ $> cp sample.env .env
...
+      <div>
+      	<h1>{this.getIntlMessage('key_name_here')}
+      </div>
+    );
+  }
+
+});
+```
+
+Once you added the mixin it will expose `getIntlMessage` method for your component to get the localized message for the given key name.
```
ScottDowne: I think what `getIntlMessage` does is probably pretty clear. It's more important that we're very clear about how to load it, so I think the less we say here the clearer it is. So maybe something like this:

"Adding the `IntlMixin` mixin exposes `getIntlMessage` to your component."

README.md
alicoding: pretty much except file name and locale in YAML must match what's in `supportedLocales`

`en-US.yaml` and in YAML `en-US` at the top.

README.md
```diff
@@ -16,3 +16,60 @@ $> cp sample.env .env
...
 $> npm start
 ```
+
+## Localization
+
+In this project we're using [React-Intl](https://github.com/yahoo/react-intl) to localize our application and YAML for translation.
+
+#### Localize a component or page
+
+To localize a component or page you have to always include IntlMixin in your class mixins for example:
```
ScottDowne: Oh, I was reading things bottom up, and didn't see this before doing the comment below.

ScottDowne: Some small changes here:

"To localize a component or page you have to include `IntlMixin` in your class `mixins`, for example"

README.md
ScottDowne: Reading this from the top down, I'm now mostly OK with this bit, I just have tiny changes now.

"Once you add the mixin it will expose `getIntlMessage` method to your component to get the localized message for the given key"

I changed added/add for/to and removed name

ScottDowne:
I have a handful of comments that might make this clearer.

alicoding:
@ScottDowne fixed all of them. Can you re-review again?

ScottDowne:
Looks good. Once Travis is done merge it!

ScottDowne:
Oh, Travis is done, so I shall merge.

PR #109, opened 2015-07-28
Change url for sequential-{locale} page to donate-{locale}

### LINKED ISSUE #100: Don't call the default donation url 'sequential'
The word 'sequential' is just an artifact from when we were testing multiple forms on BSD. In the future, our best performing donation flow may not be 'sequential' but we'll still want a URL that makes sense.

screenshot 2015-07-28 14 21 32

Our default (for EN) is https://donate.mozilla.org

For localized versions of our main form lets use:

https://donate.mozilla.org/{locale}
or
https://donate.mozilla.org/donate-{locale}

alicoding:
fixes #100


alicoding:
@ScottDowne R?

Issue #16, opened 2015-06-10
Analytics Setup and QA


adamlofting:

adamlofting:
- [ ] Decide which GA profile is going to house this data (is it shared with advocacy or not? I suggest not)
- [ ] Setup new profile + Staging
- [ ] Add GA tags to templates within code base (are their templates?)
- [ ] Add conversion tracking to thank you page to cope with PayPal & Stripe Parameters that can be returned
- [ ] Add Optimizely tags to templates
- [ ] Add Optimizely conversion goals

adamlofting:
- [ ] Send virtual page views on sequential steps
- [ ] On conversion tracking, filter out major gifts as they skew revenue data in optimization tests
- [ ] Log recurring gifts under another product name, and multiply value by 12 for GA

PR #114, opened 2015-07-29
Track sequential steps as virtual page views

adamlofting:


pages/sequential.jsx
```diff
@@ -4,6 +4,8 @@ import Header from '../components/header.jsx';
 import Link from '../components/link.jsx';
 import {IntlMixin} from 'react-intl';
 
+var ga = require('react-ga');
```
adamlofting: Should I be using `import` like above? This looks like shiny new Javascript I'm not yet familiar with :stuck_out_tongue_closed_eyes:

adamlofting:
@ScottDowne R?

(I can confirm data is arriving in GA via this running locally)

pages/sequential.jsx
alicoding: Not require really :P we're just trying to be cool haha

alicoding:
R+

adamlofting:
Completes more of #16

PR #118, opened 2015-07-29
escape an escape in a regex in a string in an attribute

ScottDowne:


ScottDowne:
Initially the double escape wasn't working for me, but it is now. Should be an OK fix I guess...

R?

alicoding:
r+

Issue #117, opened 2015-07-29
Copy across current language strings from live forms


adamlofting:
Let's get these strings into the new pages:
https://github.com/mozilla/bsd-forms-and-wrappers/tree/master/0-donation-forms/locales

adamlofting:
We have two sets of locale strings for BSD pages.
https://github.com/mozilla/bsd-forms-and-wrappers/tree/master/0-content-page/EOYFR2014-PayPal-Donate/locales

adamlofting:
For each locale, what we need to copy across is all the strings in: 
* https://github.com/mozilla/bsd-forms-and-wrappers/tree/master/0-donation-forms/locales
plus 3 strings from 
https://github.com/mozilla/bsd-forms-and-wrappers/tree/master/0-content-page/EOYFR2014-PayPal-Donate/locales

1. donate_today
2. error_amount
3. donate_now_amount

The rest are duplicates.

I'm working through this now.

PR #123, opened 2015-07-30
Import locale strings and match to yaml format

adamlofting:
I also make the keys in the yaml files more consistent (underscores, lowercase, etc)


adamlofting:
Fixes #117 

R? @alicoding

alicoding:
Thanks for doing this @Adamlofting r+

PR #127, opened 2015-07-30
Fixes #124 Localize sequential form

### LINKED ISSUE #124: Get locale strings into the site
Following on from #117 

These strings should be used throughout the pages. 

I'll work on this now.

adamlofting:
R? @alicoding

### DIFF
--- a/locales/en-US.yaml
+++ b/locales/en-US.yaml
@@ -24,9 +24,9 @@ en-US:
   credit_card_expiration_year: Credit Card Expiration Year
   email_info: We&rsquo;ll email you a receipt for your donation. We won&rsquo;t send you other email without your permission.
   pp_acknowledge: To submit this form you must agree with our privacy policy
-  privacy_policy: I&rsquo;m okay with you handling this info as you explain in your <a href=\"https://www.mozilla.org/privacy/\" target=\"_blank\">privacy policy</a>
+  privacy_policy: I&rsquo;m okay with you handling this info as you explain in your <a href="https://www.mozilla.org/privacy/" target="_blank">privacy policy</a>
   please_complete: Oops, please complete fields highlighted in red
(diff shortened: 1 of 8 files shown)


adamlofting:
Fixes #124

locales/en-US.yaml
```diff
@@ -56,3 +56,4 @@ en-US:
   continue: continue
   error_amount: Please select a value that is no less than {{ min_amount }}.
   donate_now_amount: Donate {{ currency_symbol }}{ amount } now
+  other_ways_to_give: "Other ways to give: <Link to='give-bitcoin'>Bitcoin</Link> | <a target='_blank' href='https://wiki.mozilla.org/Ways_to_Give#Check_.28via_postal_service.29'>Check</a>"
```
alicoding: I don't know if this would work? Since component in translation is not allowed by React-Intl.

locales/es.yaml
```diff
@@ -56,3 +56,4 @@ es:
   continue: continuar
   error_amount: Por favor, seleccione un valor que no sea menor que {{ min_amount }}.
   donate_now_amount: Donar {{ currency_symbol }}{ amount } ahora
+  other_ways_to_give:
```
alicoding: Don't include untranslated strings here.

alicoding:
There are few more things to fix, but almost there.

locales/es.yaml
adamlofting: My goal here was to hide this text on the localized pages (as the places they link too are not localized). Is this a bad way to do that?

locales/en-US.yaml
adamlofting: Oh yeah, let me look at this.

locales/es.yaml
alicoding: I would rather do it in the code than in this translation file since this will be display on Transifex.

locales/en-US.yaml
alicoding: @adamlofting do it similar to this way: https://github.com/mozilla/webmaker-core/blob/583f81911df650775b96b2d2246b083e929c45f5/src/pages/login/sign-in.jsx#L144-L146

PR #129, opened 2015-07-30
Missing string fixes for i18n

### LINKED ISSUE #125: Work out what to do with 'incomplete locales'
In old BSD pages, we had two sets of locale strings. See notes in #117 

The new setup builds out the static pages based on the locale files in `/locales/` but we have a bunch of locales (e.g. JP) where we only have a smaller subset of strings which were used for the simple PayPal pages.

If we import these partial locale files, the build will break at the moment.

Let's work out a plan for moving forward with this.

alicoding:
Fixes #125

If we have partial translation in some languages this will help us solve
the problem by filling the default locale values.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -6,6 +6,7 @@ var Router = require('react-router');
 var routes = require('./components/routes.jsx');
 var paths = require('./scripts/paths.js');
 var currencies = require('./scripts/currencies.js');
+var englishStrings = require(path.join(__dirname, 'locales/en-US.json'));
 
 module.exports = {
   entry: "./components/client.jsx",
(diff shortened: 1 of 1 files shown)


alicoding:
@adamlofting r?

adamlofting:
I'm on calls for the rest of the day, so bouncing R? to @ScottDowne in case you have time to look at this before me.

webpack.config.js
```diff
@@ -6,6 +6,7 @@ var Router = require('react-router');
 var routes = require('./components/routes.jsx');
 var paths = require('./scripts/paths.js');
 var currencies = require('./scripts/currencies.js');
+var englishStrings = require(path.join(__dirname, 'locales/en-US.json'));
```
ScottDowne: What's the purpose of defining this way up here? Or was it just cleaning things up/as good as anything?

ScottDowne: I think I see the idea here.

ScottDowne:
So I'm a bit concerned with building the whole app for locales that only need 10% of the app.

I'm still going to push back on that, but it also occurred to me that not failing on missing strings is probably handy, and that's all this is doing.

So R+ on this because they are two problems, for now.

ScottDowne:
Once we have all the locale files for the paypal currency pages, if the build step is noticeably slower, we'll need to optimize.

adamlofting:
> So I'm a bit concerned with building the whole app for locales that only need 10% of the app.

> I'm still going to push back on that, but it also occurred to me that not failing on missing strings is probably handy, and that's all this is doing.

Agreed. I think we might need to modify the build step so that we can say something like:

* Build `form-a` in locales `[ a, b, c, d ]`
* Build `form-b` in locales `[ c, d, e, x, y ]`

E.g. Build sequential form in some locales, and build simple paypal form in others.

But we should write this in a way that works when we're testing new landing pages in multiple locales.

adamlofting:
Just realised this is the PR and not the ticket. I'll file those comments in a new ticket to pick up later.

PR #132, opened 2015-07-30
allow server to start and stop via nodejs and add a test that does th…

ScottDowne:
…is so travis can catch obvious server startup errors


ScottDowne:
To test this, require in nonsense, like `require('nonsense');` in server.js.

Push that up to travis, and it should fail.

ScottDowne:
r?

alicoding:
do you still need to run two commands to run the test here?

ScottDowne:
no, the selenium test now runs itself, but that's it.

What I did fix is selenium runs the server, and travis runs the server.

I didn't do selenium tests run via travis, and selenium stopping itself.

alicoding:
r+

PR #137, opened 2015-07-30
Fix issue101 - Localize Thank you page

alicoding:
This patch will only make sure to display the form for English version only.


alicoding:
@ScottDowne R?

components/form.jsx
```diff
@@ -0,0 +1,72 @@
...
+	    }
+	  });
+	},
+	render: function() {
+		return (
+			<div>
+				<div className="form-wrapper container">
+				  <div className="wrap">
+				    <div className="row">
+				      <h2>Sign up for email updates</h2>
```
ScottDowne: What's with the whitespace?

components/form.jsx
```diff
@@ -0,0 +1,72 @@
+var React = require('react');
```
ScottDowne: Can we call this signup.jsx instead of form.jsx?

ScottDowne:
Just change the name of form.jsx and hit merge.

PR #148, opened 2015-07-31
display hidden stripe notice once credit card is selected

ScottDowne:


ScottDowne:
@alicoding Another review? Turns out we moved this around a bit.

It's now the bottom link in the grey area, instead of the middle. It also shows up only in step 2 after cc is clicked.

alicoding:
r+

PR #151, opened 2015-07-31
Sets up GA to track recurring donations distinct from one-off gifts

adamlofting:


components/simple-paypal.jsx
```diff
@@ -177,7 +177,7 @@ var simplePaypal = React.createClass({
           <input type="hidden" name="cmd" value="_xclick-subscriptions"/>
           <input type="hidden" name="business" value="44ZHAVWJHTK2N"/>
           <input type="hidden" name="lc" value={this.props.paypalLocal}/>
-          <input type="hidden" name="item_name" value="Mozilla Foundation Recurring Donation"/>
+          <input type="hidden" name="item_name" value="Mozilla Foundation Monthly Donation"/>
```
adamlofting: I thought we should be specific in the name about how often the donation occurs.

public/js/analytics.js
```diff
@@ -32,15 +43,37 @@
   var str_tx_re = /str_id=(\w+)/;
   var str_amt_re = /str_amount=([\w\.]+)/;
   var str_cc_re = /str_currency=(\w+)/;
+  var str_frequency_re = /str_frequency=(\w+)/;
```
adamlofting: We're not actually passing this param through yet for Stripe recurring, but the GA code is ready to go when Stripe is.

adamlofting:
This finishes last item on #16

components/simple-paypal.jsx
alicoding: I like Monthly vs Recurring here. :+1:

alicoding:
r+

PR #154, opened 2015-07-31
no usage of global.window check, because files are only run in a cont…

ScottDowne:
…ext in which they can run successfully


ScottDowne:
Adding comment with explaining this better very soon...

ScottDowne:
Experimenting, bear with me.

ScottDowne:
The idea I am experimenting with is this:

Client side only:
public/js/*
Client.jsx
componentDidMount

Universal:
scripts/*
jsx files minus Client.jsx

Server side only:
Everything else

ScottDowne:
Actually, looks like whatever is in scripts now is very different than what it was. It's basically server side only now plus some simple require json data files.

That makes things even easier!

ScottDowne:
@alicoding Ha, I want you to try your hardest to shoot this down. I'll accept nothing less.

components/client.jsx
```diff
@@ -3,7 +3,7 @@ import assign from 'react/lib/Object.assign';
 import Router from 'react-router';
 import currencies from '../scripts/currencies.js';
 import routes from './routes.jsx';
-import {intlDataFor, intlData} from '../scripts/i18n';
+import {intlDataFor, intlData} from '../public/js/i18n.js';
```
alicoding: Unnecessary .js here but just nit picking lol

components/routes.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { Route } from 'react-router';
-import pages from '../scripts/pages.js';
+import pages from '../data/pages.js';
```
alicoding: I like that we abstract things to its own descriptive directory here

locales/index.js
```diff
@@ -0,0 +1,11 @@
+function getLocales(arr) {
+  return arr.map(function(item) {
+    return item.replace("./", "").replace(".json", "");
+  });
+}
+
+var reqF = require('enhanced-require');
+var req = reqF(module);
```
alicoding: Not sure why I didn't thought of this lol! It can pretty much run both client and server with this module.

alicoding:
I still don't know what makes your changes work here, but kinda get the idea I guess... As long as this is working I'm happier than having to use that global.window, so r+

PR #179, opened 2015-08-06
Fix double slash

alicoding:
This was fixed by Babel, so now we don't have to escape it anymore :)


mmmavis:
R+ awesomeeeeeeeeee

PR #202, opened 2015-08-10
simplified form submission for single page, also made sequential stil…

ScottDowne:
…l work

### DIFF
--- a/components/amount-buttons.jsx
+++ b/components/amount-buttons.jsx
@@ -2,12 +2,18 @@ import React from 'react';
 
 var AmountButtons = React.createClass({
   mixins: [require('react-intl').IntlMixin],
+  getInitialState: function() {
+    return {
+      otherValue: "0"
+    };
+  },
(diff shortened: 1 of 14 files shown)


components/amount-buttons.jsx
```diff
@@ -110,9 +139,9 @@ var AmountButtons = React.createClass({
           </div>
           <div className="two-third">
             <div id="amount-other-container">
-              <input type="radio" name="donation_amount" value="other" id="amount-other" data-parsley-multiple="donation_amount" data-parsley-group="page-1"/>
+              <input type="radio" name="donation_amount" value={this.state.otherValue} id="amount-other" data-parsley-multiple="donation_amount" data-parsley-group="page-1"/>
```
alicoding: I start to wonder if you should change this to `linkState` instead and whenever or wherever you're trying to use it you can always pull it from the `state` object instead.

ScottDowne:
This landed in three other chunks.

Issue #138, opened 2015-07-31
Move domain names into environment vars


adamlofting:
Let's dig around the repo and move some of these into environment vars:

References to our staging / prod domains
* https://github.com/mozilla/donate.mozilla.org/blob/master/server.js#L65

References to Paypal sandbox
* https://github.com/mozilla/donate.mozilla.org/blob/master/server.js#L80

And then let's document these vars in the README

The existing values are good defaults for developers to work with for local dev.

ScottDowne:
This is all I could find that is left: https://github.com/mozilla/donate.mozilla.org/commit/71ee920214933f0d60772108ff939ad497764652

adamlofting:
There's a couple in here: 
* https://github.com/mozilla/donate.mozilla.org/blob/master/components/simple-paypal.jsx#L162

And a coinbase one here:
* https://github.com/mozilla/donate.mozilla.org/blob/master/pages/give-bitcoin.jsx

And this for email sign-up here (we don't have staging ATM, but will do once we move to Basket):
* https://github.com/mozilla/donate.mozilla.org/blob/master/routes/index.js#L9

ScottDowne:
Another pass: https://github.com/mozilla/donate.mozilla.org/pull/320

I think that got all of them.

PR #209, opened 2015-08-11
boil down routes

ScottDowne:


ScottDowne:
r?

ScottDowne:
@jbuck After this lands, I figure if you do one error handle, I'll copy that pattern and do the rest.

pages/sequential.jsx
```diff
@@ -537,8 +537,8 @@ var Sequential = React.createClass({
                     </div>
                   </div>
                   <div className="row" id="payment-type-row">
-                    <PaypalButton />
```
jbuck: Should this change be in this patch?

routes/index.js
```diff
@@ -0,0 +1,104 @@
+var httpRequest = require('request');
+var stripe = require("./stripe");
+var paypal = require("./paypal");
+
+var routes = {
+  "signup": function(request, reply) {
+    var transaction = request.payload || {};
+    httpRequest({
+      url: 'http://donate-mozilla.herokuapp.com/thank-you/',
```
jbuck: The host is hard-coded and I think the path is wrong here too

alicoding: https://github.com/mozilla/donate.mozilla.org/issues/138

jbuck:
@ScottDowne r+

PR #222, opened 2015-08-12
Fix issue #103 - only include analytics.js on certain pages

### LINKED ISSUE #103: Only load js/analytics.js on thank you pages
Currently, js/analytics.js is loaded on all pages.

But we only need to call this on
`/thank-you*`
and
`/share*`
pages

This results in one less http request, and a few less regex expressions to run on the rest of our forms.

Ideally, we could mark a page as being of type 'thank-you' so that if we create multiple thank you pages for testing, we don't forget to add this tracking in the future.

alicoding:
closes #103

### DIFF
--- a/pages/index.jsx
+++ b/pages/index.jsx
@@ -14,7 +14,6 @@ var Index = React.createClass({
           <script src="https://js.stripe.com/v1/"></script>
           <script src="/js/stripe.js"></script>
           <script src="/js/ga.js"></script>
-          <script src="/js/analytics.js"></script>
           <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Open+Sans:400,300,300italic"/>
           <link href="//maxcdn.bootstrapcdn.com/font-awesome/4.4.0/css/font-awesome.min.css" rel="stylesheet"/>
           <link rel="icon" href="/images/favicon.ico" type="image/x-icon"/>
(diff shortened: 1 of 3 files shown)


alicoding:
@ScottDowne R?

pages/share.jsx
```diff
@@ -2,9 +2,13 @@ import React  from 'react';
 import Footer from '../components/footer.jsx';
 import Header from '../components/header.jsx';
 import Social from '../components/social.jsx';
+import analytics from '../public/js/analytics.js';
 
 var ThankYou = React.createClass({
   mixins: [require('react-intl').IntlMixin],
+  componentDidMount: function() {
+    analytics();
```
ScottDowne: I am pretty sure we don't want this here.

1. It'll double count locales with share in the thank you page.
2. Share itself isn't a page that has transaction data we need to track.

ScottDowne:
Just the one comment.

pages/share.jsx
alicoding: Based on this https://github.com/mozilla/donate.mozilla.org/issues/103 not sure if I misunderstood something from there?

ScottDowne: Yeah, and share is not the thank you page.

ScottDowne: Oh, I see, adam is asking explicitly for share to have it... odd.

PR #238, opened 2015-08-14
Stripe recurring parameters for analytics - fixes #119

### LINKED ISSUE #119: metrics on recurring payments
Not sure how we handle this.

The transaction isn't what we expect for recurring like it is for a single donation.

So the metrics are failing for recurring.

adamlofting:
Fixes #119 

R? @ScottDowne


alicoding:
r+

PR #245, opened 2015-08-14
update button based on if it is monthly or single

ScottDowne:


ScottDowne:
r?

pages/sequential.jsx
```diff
@@ -151,14 +151,27 @@ var Sequential = React.createClass({
     var updateDonateButtonText = function(amountSelected) {
       var buttonText;
       var locale = "US";
+      var recurring = $('[name="recurring_acknowledge"]:checked').val() === '1';
       if (amountSelected && locale === "US") {
-        buttonText = "Donate $" + amountSelected + " now";
+        if (recurring) {
+          buttonText = "Donate $" + amountSelected + " monthly";
```
alicoding: let's not do this. we want to give control for localizer to change the order around. You are forcing them to use one strict rule for the formatting here.

alicoding:
r-

PR #246, opened 2015-08-14
Fix travis build

alicoding:
Travis was failing because of some linting issue.

I have to create a common API file to share these methods across which
is also a good practice in case we've change one of the method in
sequential file it wouldn't cause bug in the other.


alicoding:
@ScottDowne r?

PR #248, opened 2015-08-14
remove hard coded paypal sandbox link

ScottDowne:


ScottDowne:
r?

routes/index.js
```diff
@@ -60,7 +60,7 @@ var routes = {
         if (err) {
           return console.error('donation failed:', err);
         }
-        reply.redirect("https://www.sandbox.paypal.com/cgi-bin/webscr?cmd=_express-checkout&useraction=commit&token=" + charge.TOKEN);
+        reply.redirect(process.env.PAYPAY_ENDPOINT + "/cgi-bin/webscr?cmd=_express-checkout&useraction=commit&token=" + charge.TOKEN);
```
alicoding: Can you set it to a variable and then use it instead? Also fallback to something in case I don't have this in my environment?

adamlofting: Also, is `PAYPAY` a typo?

PR #250, opened 2015-08-14
adding stripe widget to single page form

ScottDowne:


alicoding:
components/stripe-button.jsx: line 13, col 19, 'StripeCheckout' is not defined.

Issue #81, opened 2015-07-20
Configure HTTP header security


adamlofting:
To help mitigate risks of XSS / Clickjacking

PR #255, opened 2015-08-16
Add HTTP header security & Content Security Policy

jbuck:


jbuck:
Fix #81

server.js
```diff
@@ -12,6 +10,14 @@ server.connection({
   port: process.env.PORT
 });
 
+var securityConfig = {
+  hsts: process.env.HSTS_ENABLED === "true" ? 15768000 : false,
```
alicoding: Do we need this to be part of the sample.env as well?

alicoding:
R+ from me with one question left

jbuck:
Hm, I think I'm going to wait until tomorrow so this can get a more thorough review... I bet I broke some other pages. We should have some sort of a QA script I can run through first!

alicoding:
1. make sure it is the latest commit as well
check out jbuck's patch
2. we want to nuke any precompiled version of anything inside that public/ directory and then reset the branch
rm -rf public/ && git checkout .
3. maybe there is an update to the package.json
npm install
4. now run the server
npm start
5. visit the site
http://localhost:3000/
6. test donation manually for both credit card and paypal; monthly and one-time payment
7. for credit card use this number for test: 4242424242424242 - month and date any in the future and for CVC any 3 digits
8. fill in the form don't leave blanks, or leave some blanks to see any errors
9. also check the button if the value and the text reads correctly based on your amount selection
10. after submitted you should see thank you page 
11. check the url if it is similar or not missing any query
http://localhost:3000/thank-you/?payment=Stripe&str_amount=2000&str_currency=usd&str_id=ch_16ao8MKvjrcBLqIANpqlz7is&str_frequency=one-time
12. if you submit the form you should see /share page
http://localhost:3000/share/

r- because I'm stuck on paypal part and the error I'm getting is:
```
donation failed: [Error: options.uri is a required argument]
```

jbuck:
@alicoding Okay, try this patch again. It should be working now that I've rebased it on master!

adamlofting:
Let's enable [review apps](https://devcenter.heroku.com/articles/github-integration-review-apps) to turn this PR into a demo app which I can then test in Optimizely.

alicoding:
:+1: with Adam about review apps. I will test this again on my machine anyway so that I can confirm about that error if it's fixed.

alicoding:
@jbuck did you rebased with master? I'm still getting this error when trying to go through PayPal's option here.
```
donation failed: [Error: options.uri is a required argument]
```

PR #263, opened 2015-08-19
Cloudformation deploy tool

jbuck:


jbuck:
@alicoding Can you help me figure out why I'm getting this linting error? Line 14 is the first point where I define `name` so I don't understand why I'm getting this error

adamlofting:
I don't have time to check this out, get creds, run it etc today.

But from a visual review this looks good, I've also seen it running on your machine, the docs make sense, and it doesn't affect the user facing site.

So, unless you really want a really good review, I'm saying R+

PR #265, opened 2015-08-20
components for fun and profit

ScottDowne:


alicoding:
This patch is huge

ScottDowne:
Yeah but once it's done, sequential.jsx can be deleted, and likely shared.less too.

Oh right, this is a WIP. I still need to do a bunch of things...

components/address-input.jsx
```diff
@@ -0,0 +1,628 @@
+var React = require('react');
```
alicoding: Import?

components/amount-buttons.jsx
```diff
@@ -4,17 +4,68 @@ var AmountButtons = React.createClass({
...
+    this.setAmount();
+  },
+  setAmount: function() {
+    this.setState({
+      valid: true
+    });
+    this.props.onChange(this.props.name, this);
+  },
+  otherRadioClick: function() {
+    $("#amount-other-input").focus();
```
alicoding: Don't use jquery and use either native or react built-in query selector

alicoding:
rebase and fix that jquery stuff

ScottDowne:
@alicoding rebased and ready for r again

PR #267, opened 2015-08-20
Switch from JSXHint to ESLint

alicoding:
JSXHint is deprecated and it was a bit buggy which causes a lot of problem with what we're using here. I'm switching it to ESLint which support both JS and JSX.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,38 @@
+{
+    "rules": {
+        "indent": [
+            2,
+            2
+        ],
+        "quotes": [
+            2,
+            "single"
+        ],
+        "linebreak-style": [
+            2,
+            "unix"
+        ],
+        "semi": [
+            2,
+            "always"
+        ],
+        "no-console": 0
+    },
+    "env": {
+        "es6": true,
+        "node": true,
+        "browser": true
+    },
+    "globals": {
+      "$": true
+    },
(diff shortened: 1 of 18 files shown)


alicoding:
@ScottDowne R?

.eslintrc
```diff
@@ -0,0 +1,38 @@
+{
+    "rules": {
+        "indent": [
+            2,
+            2
+        ],
+        "quotes": [
+            2,
+            "single"
```
ScottDowne: Hilarious that the file that says we use single quotes is in a file that's all double.

ScottDowne: Same with some other files we have that I guess we don't bother linting, like package.json.

alicoding: What other file? Also we only test .js and .jsx, so any JSON is ignored

alicoding:
it looks like there are many other .jsx file that is not currently linting, but I will try to enable this tomorrow.

.eslintrc
ScottDowne: I just find it amusing. Nothing I expect fixed.

alicoding:
@ScottDowne not sure if you said it is good to merge here? Can you confirm if you have time if I can merge this patch?

ScottDowne:
@alicoding I just wanted to run it first. Good to go now.

PR #273, opened 2015-08-21
adding docs for server client relation, and an empty file for compone…

ScottDowne:
…nt docs

### DIFF
--- a/docs/Client_Server.md
+++ b/docs/Client_Server.md
@@ -1,0 +1,39 @@
+# Client and Server
+
+## Precompile
+
+Server side we precompile all our React components into static HTML.
+
+Use the `npm install` command to trigger or update the precompile.
+
+### Webpack
+
+Webpack creates static HTML pages from a list of react routes.
+
+First it gets an array of all our paths from `/scripts/paths.js`.
+
+`/scripts/paths.js` takes basic pages from `/data/pages.js`.
+
+`/data/pages.js` is a simple list of pages, it doesn't deal with locale and currency pages. `/scripts/paths.js` then compiles a list of all possible paths based on our pages.
+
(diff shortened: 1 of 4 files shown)


docs/Client_Server.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Precompile
+
+Server side we precompile all our React components into static HTML.
+
+Use the `npm install` command to trigger or update the precompile.
+
+### Webpack
+
+Webpack creates static HTML pages from a list of react routes.
```
alicoding: I think you might want to be specific here and mention about Webpack plugin because Webpack doesn't do this by default, so it might confused people.

alicoding: Ah I see you mentioned that in line 19, but I think you might want to talk about it first here.

alicoding:
Initial comment on this READMEs. I feel like it is good for your presentation, but a bit confusing to just a doc for people to read to understand what's going on here.

docs/Client_Server.md
ScottDowne: It does sound like I am saying this is a feature of webpack.

Something like "In our webpack.config we create static HTML pages from a list of react routes."

I'm not sure I want to mention the plugin yet, In that line, I'm trying to stick to what we're doing, not how.

alicoding: Yeah saying that the webpack.config.js is sound much better than just webpack

ScottDowne:
Can you be more specific? What you expect from them? If it's confusing here, it'll be confusing in a presentation. I also by no means meant for this to be good for a presentation, it was meant to be docs, so that means I'm totally off the mark.

docs/Client_Server.md
ScottDowne: :D agreed!

ScottDowne:
@alicoding r?

alicoding:
r+

PR #274, opened 2015-08-21
Docs about our use of GA

adamlofting:
Just some docs :smile: 

R?

### DIFF
--- a/docs/Metrics.md
+++ b/docs/Metrics.md
@@ -7,9 +7,9 @@
 
 ## Instrumentation (Google Analytics)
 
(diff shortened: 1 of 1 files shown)


docs/Metrics.md
```diff
@@ -1,11 +1,48 @@
...
+3. [Exchange Rates](#exchange-rates)
+
+
 ## Instrumentation (Google Analytics)
 
-* TODO: add content about GA
-  * convesion tracking
-  * virtual views
-  * react-ga lib
+We load Google Analytics on all of our fundraising web pages, via [index.jsx](https://github.com/mozilla/donate.mozilla.org/blob/master/pages/index.jsx#L16). This is the standard Universal Analytics tracking tag, which gives us default data about page views [code](https://github.com/mozilla/donate.mozilla.org/blob/master/public/js/ga.js).
```
alicoding: I don't know if pointing to a specific line is a good idea here? Might be better to include example instead? Because it often get moves around and then pointing to the wrong line?

alicoding:
r+ with one fix

docs/Metrics.md
adamlofting: Good point. I'll fix this up now.

PR #275, opened 2015-08-22
Fix #257 - Re-Structure URL with language code

### LINKED ISSUE #257: Re-order URL for Locales

alicoding:
This patch restructure the URL with language code, for example:

We use to have this URL for sequential page: `/sequential-de`
This patch changed this to be `/de/sequential`.

By design doing it this way is better for both technical and user point
of view.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,9 @@ public/*/index.html
 public/main.js
 public/exchange-rates/latest.json
 locales/*.json
+public/de/
+public/en-US/
+public/es/
+public/fr/
+public/id/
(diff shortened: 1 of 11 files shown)


alicoding:
@ScottDowne You would like this more than when I spoke to you because I don't have to check for `window` at all :P

alicoding:
Just ignore this failing test. I will fix any linting issue once that ESLint patch is land @ScottDowne

alicoding:
@ScottDowne I've fixed linting problems in this patch now, so ready for review once you have time.

PR #280, opened 2015-08-25
Adding cookieDomain before Optimizely

adamlofting:
This is WIP - just opening PR so it's easier to talk through.

### DIFF
--- a/pages/index.jsx
+++ b/pages/index.jsx
@@ -1,10 +1,20 @@
 var React = require('react');
 
+var Habitat = require('habitat');
+Habitat.load();
+var env = new Habitat();
+
+var cookieDomain = env.get('FULL_SUBDOMAIN_FOR_COOKIE');
+
 var Index = React.createClass({
   render: function() {
     return (
       <html>
         <head>
+          <script>
+            window['optimizely'] = window['optimizely'] || [];
+            window['optimizely'].push(["setCookieDomain", {{cookieDomain}}]);
+          </script>
           <script src="https://cdn.optimizely.com/js/206878104.js"></script>
           <meta charSet="UTF-8"/>
           <meta name="viewport" content="width=device-width, initial-scale=1"/>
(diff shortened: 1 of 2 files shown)


pages/index.jsx
```diff
@@ -1,10 +1,20 @@
...
+
 var Index = React.createClass({
   render: function() {
     return (
       <html>
         <head>
+          <script>
+            window['optimizely'] = window['optimizely'] || [];
+            window['optimizely'].push(["setCookieDomain", {{cookieDomain}}]);
+          </script>
```
adamlofting: This won't work with React, as it gets rendered like this:

```js
<script>window[&#x27;optimizely&#x27;] = window[&#x27;optimizely&#x27;] || []; window[&#x27;optimizely&#x27;].push([&quot;setCookieDomain&quot;, localhost]);</script>
```

What's the best way to add this in React?

@ScottDowne or @alicoding ?

jbuck: Out of curiosity, why do you need to set the subdomain? Isn't it automatically scoped to the correct domain?

adamlofting: It works by default, but we're being extra careful:
https://bugzilla.mozilla.org/show_bug.cgi?id=1125995

jbuck: Ahhh, I see. So optimizely sets the cookie domain to mozilla.org by default. laaaaaaame

alicoding: Can you also move it to `componentDidMount` instead?

pages/index.jsx
```diff
@@ -1,10 +1,20 @@
 var React = require('react');
 
+var Habitat = require('habitat');
+Habitat.load();
```
ScottDowne: Woah, this works client side?

adamlofting: I assumed this index.jsx file was just being run in the pre-compile. And we'd end up with static HTML... but I might have missed something important here.

alicoding: I would say use this way instead? https://github.com/mozilla/id.webmaker.org/blob/develop/webpack.config.js#L42-L50

ScottDowne:
@adamlofting this still WIP?

adamlofting:
> @adamlofting this still WIP?

Yep, I started on some fixes to this but didn't get time to finish it today.

adamlofting:
I ended up taking a different approach, so will open a new PR.

PR #281, opened 2015-08-25
Fix bitcoin inaccessible

alicoding:
I've broke the access to that page when you click on the link on home
page. This should fix it now and should also remove those warnings on
startup of the app.


alicoding:
@ScottDowne R?

PR #307, opened 2015-08-26
Remove hash from URL for paging

alicoding:
I don't think that part is even working now, and we are relying on state
to keep track of our paging, so I don't see a reason to keep this.

Is there a good reason that we want to keep this @ScottDowne?

### DIFF
--- a/components/amount-buttons.jsx
+++ b/components/amount-buttons.jsx
@@ -114,46 +114,6 @@ var AmountButtons = React.createClass({
         event.preventDefault();
       }
     });
-
-    // ***********************************************
-    // Preselect dollar amount
-    // ***********************************************
-    function processAmountHash(winLocationHash) {
-      var amtRegex = /#amount-([\w\.]+)/;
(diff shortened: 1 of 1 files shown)


adamlofting:
I think it's fine to delete this existing code if it isn't working any more, but we will need to add this functionality back in to work with the new forms.

It doens't have to be a hash though, it can just be URL parameters.

We use this functionality for everyone who comes from the snippet as they have preselected an amount we want to take the straight to page 2.

ScottDowne:
@adamlofting We should remove the page-1 bits, as that won't work well with loosely coupled forms.

We should do something like (not in this ticket) this: "?amount=5&type=cc" in that case, it'll need to be smart enough to navigate to the proper page if there are pages.

ScottDowne:
Actually, now that I think of it, page should be explicit. A component shouldn't need to care what page it's in.

I'm fine with this. It does remove pretty busted functionality for amount-n and cc-amount-n but we should redo that better anyway.

PR #314, opened 2015-08-26
Fix issue #311 - Intl is undefined for Safari

### LINKED ISSUE #311: Our page doesn't work on Safari
```
[Error] ReferenceError: Can't find variable: Intl
```

alicoding:
Adding polyfill for Safari for Intl and navigator.languages

Closes #311


alicoding:
@ScottDowne and @jbuck R?

jbuck:
I don't think we should land this patch as-is.

https://github.com/andyearnshaw/Intl.js/#intljs-and-browserifywebpack indicates that we'll need to load the locale data as well. We should be loading the correct locale data too. And the accept-language polyfill is 100% not ready for production use

ScottDowne:
Status on this?

alicoding:
I need to update webpack to also include other language pack.

alicoding:
@jbuck R?

alicoding:
r+ from the call will file an issue if we want to optimize.

PR #315, opened 2015-08-26
adding in stripe checkout thingy

ScottDowne:


ScottDowne:
Also, right now I'm saving a copy of the stripe checkout source:  public/js/stripe-checkout.js probably not ideal.

Also, I need to look into definePlugin to handle getting env to client side. https://gist.github.com/haf/f671f1113d2c5dead5a7#file-gistfile1-txt-L110

ScottDowne:
R?

mixins/form.jsx
```diff
@@ -61,54 +61,86 @@ module.exports = {
...
+    var success = this.stripeSuccess;
+    var valid = this.validateProps(validate);
+    var submitProps= {};
+    if (!valid) {
+      return;
     }
+    submitProps = this.buildProps(props);
+    var handler = StripeCheckout.configure({
+      // Need to get this from .env
+      key: 'pk_test_BZ0QTIwe7BVAk1ZDxOgWZ9Z6',
```
alicoding: get it from `.env` instead

mixins/form.jsx
```diff
@@ -61,54 +61,86 @@ module.exports = {
...
+    submitProps = this.buildProps(props);
+    Stripe.card.createToken({
+      number: submitProps.cardNumber,
+      cvc: submitProps.cvc,
+      exp_month: submitProps.expMonth,
+      exp_year: submitProps.expYear
+    }, function(status, response) {
+      if (response.error) {
+        // handle all error cases?
+        console.log(response.error);
```
alicoding: How are we planning to handle this error?

alicoding:
r+ with that changes and one question to be answered

PR #317, opened 2015-08-26
fix virtual page tracking

ScottDowne:


ScottDowne:
R?

I didn't change the way GA expects the tracking info, I think that's a question for Adam in another ticket. I'll file that.

alicoding:
r+

PR #319, opened 2015-08-26
adding back small print bit below form

ScottDowne:


ScottDowne:
R?

Should be pretty straight forward.

components/small-print.jsx
```diff
@@ -0,0 +1,35 @@
...
+  render: function() {
+    var stripeNotice = "stripe-notice";
+    if (!this.props.stripeNotice) {
+      stripeNotice += " hidden";
+    }
+    return (
+      <div className="row disclaimers">
+        <p className="other_ways_to_give">
+          <small>
+            Other ways to give: <Link to='give-bitcoin'>Bitcoin</Link> |
```
alicoding: Can you localize this one as well? Not sure how we miss that.

alicoding:
r+ with that one fix

PR #322, opened 2015-08-26
add amount preselect and add preset amounts

ScottDowne:


ScottDowne:
R?

ScottDowne:
This does both presets and preselect value.

You can use it like so: ?amount=101?presets=1,2,101,102

It *only* accepts four preset values. Any more or any less, and it falls back to default.

components/amount-buttons.jsx
```diff
@@ -1,47 +1,68 @@
...
+var AmountButton = React.createClass({
+  render: function() {
+    var checked = false;
+    if (this.props.value === this.props.amount) {
+      checked = true;
+    }
+    return (
+      <div className="third">
+        <input onChange={this.props.onChange} checked={checked} type="radio" name="donation_amount" value={this.props.value} id={"amount-" + this.props.value}/>
+        <label htmlFor={"amount-" + this.props.value} className="large-label-size">${this.props.value}</label>
```
alicoding: Let's use formatjs (react-intl) to handle currency display here. `formatNumber()`

ScottDowne: works for me.

alicoding:
```
Warning: Failed propType: You provided a `checked` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultChecked`. Otherwise, set either `onChange` or `readOnly`. Check the render method of `AmountOtherButton`. main.js:1650:8
Warning: Failed propType: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`. Check the render method of `AmountOtherButton`.
```

alicoding:
r+ with those fixes or flag me again if you want to make sure

components/amount-buttons.jsx
```diff
@@ -1,50 +1,103 @@
...
+          />
+        </label>
+      </div>
+    );
+  }
+});
+
+var AmountOtherButton = React.createClass({
+  onRadioClick: function() {
+    document.querySelector("#amount-other-input").focus();
```
k88hudson: Just a side note, better to use a `ref` here, or if there's some reason you can't,

```
this.getDOMNode().querySelector('...');
```

PR #326, opened 2015-08-27
Switch from .po to .properties

alicoding:
Since we are relying on IDs or key=value for our app to consume. Using
.po kind of complicate things for us. Fortunately, Pontoon supports
.properties which is also using key=value concept, so we should use this
format instead.


alicoding:
@ScottDowne R?

jbuck:
@alicoding r+

PR #328, opened 2015-08-27
Fix issue #316 - Thank-thank page broken

### LINKED ISSUE #316: Looks like thank you page is now the share page
It actually looks more like it's not properly detecting the locale, so en-us and de are thought to be a language we don't support email for right now, so it displays share instead of signup.

alicoding:
Closes #316

This should now display signup form for 'de' and 'en-US'


alicoding:
@ScottDowne R?

PR #331, opened 2015-08-27
This patch removes all currency from path and

alicoding:
And allow us to do `?currency=usd`

### DIFF
--- a/components/client.jsx
+++ b/components/client.jsx
@@ -18,13 +18,11 @@ function redirectTo(pathname, query, Handler) {
 }
 
 Router.run(routes, Router.HistoryLocation, function (Handler, state) {
-  var values = {};
   var pathname = state.query.pathname || state.pathname;
   var lang = i18n.isSupportedLanguage(i18n.currentLanguage) ? i18n.currentLanguage : i18n.defaultLang;
   var query = state.query.pathname ? {} : state.query;
-  if(state.params.currency) {
-    values = currencies[state.params.currency];
-  }
+
+  var values = query.currency && currencies[query.currency] ? currencies[query.currency] : currencies['usd'];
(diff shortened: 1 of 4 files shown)


alicoding:
@ScottDowne R?

components/client.jsx
```diff
@@ -22,9 +22,7 @@ Router.run(routes, Router.HistoryLocation, function (Handler, state) {
   var pathname = state.query.pathname || state.pathname;
   var lang = i18n.isSupportedLanguage(i18n.currentLanguage) ? i18n.currentLanguage : i18n.defaultLang;
   var query = state.query.pathname ? {} : state.query;
-  if(state.params.currency) {
-    values = currencies[state.params.currency];
-  }
+  values = query.currency && currencies[query.currency] ? currencies[query.currency] : currencies['usd'];
```
ScottDowne: I think now that this isn't an if, you can do `var values = ` and remove the `var values = {}` at the top.

alicoding: I think you still need to check if that requested currency exist

ScottDowne: Totally, you leave in the check. It would look something like this:

`var values = query.currency && currencies[query.currency] ? currencies[query.currency] : currencies['usd'];`

Then remote the `var values = {};` initialization at the top.

alicoding: Why doing var values twice?

ScottDowne: You remove the first declare, it's no longer needed.

ScottDowne: I had a typo, fixed. I typed remote when I meant remove.

alicoding: Well if you want to do value.queryString you still need to make an object first though?

PR #356, opened 2015-09-02
Localize bitcoin button

alicoding:
We need to make sure it is localizable, but it was an image, so I turned
it to a button that they use in coinbase


alicoding:
@ScottDowne R?

alicoding:
9feecdfc94c312ec53bbe8e55da63ef5ecab53da151d79603af32fccb59cd713

Issue #334, opened 2015-08-28
Put ENG strings in correct format for L10N


edrushka:
Hey @alicoding. Can you please put the English strings from this doc https://docs.google.com/document/d/1DMILjOMor2RIxJPC2uJCefUZ6vsuL2G19N0faVfUT34/edit# into the appropriate format for us to give to the L10N team? 

Hopefully this doc will be done by EOD friday Aug 28. Keep your eye on #278.

Adding this to the Sept 11 milestone but it'd be great if you could tackle as early as possible in the heartbeat so the L10N crew can get to work. Thanks!

alicoding:
Closing this ticket :) only FAQ left to do, but lets file another ticket to get that sorted out after we rewrite it in a generic way.

PR #361, opened 2015-09-03
Fix issue334 - part 01

alicoding:
There are more to do from issue #334 but I think I want to get something merged first because I need to talk to @valianttry regarding the FAQ


alicoding:
@ScottDowne R?

PR #364, opened 2015-09-03
adding errors for stripe errors, error page navigation, and waiting s…

ScottDowne:
…pinner on the donate now button


ScottDowne:
r?

components/credit-card-info.jsx
```diff
@@ -125,6 +142,17 @@ var CreditCardInfo = React.createClass({
             </div>
           </div>
         </div>
+        <div className={errorMessageClassName}>
```
alicoding: maybe we should have error component now? :P

PR #367, opened 2015-09-03
handle currency symbol in the other donation amount box

ScottDowne:


ScottDowne:
R? I wanted to do this before the currency dropdown, to sort of single display currency codes in the right amount of space. I figured I needed to do it with the other box too. I suspect I'll need to make a dropdown out of the currencies.js data.

alicoding:
R+

PR #374, opened 2015-09-04
currency changing dropdown

ScottDowne:


ScottDowne:
R?

pages/sequential.jsx
```diff
@@ -3,6 +3,7 @@ import Footer from '../components/footer.jsx';
 import Header from '../components/header.jsx';
 import SmallPrint from '../components/small-print.jsx';
 import SectionHeading from '../components/section-heading.jsx';
+import CurrencyDrop from '../components/currency-dropdown.jsx';
```
alicoding: Can you rename this to `CurrencyDropDown` instead?

pages/single-form.jsx
```diff
@@ -3,6 +3,7 @@ import Footer from '../components/footer.jsx';
 import Header from '../components/header.jsx';
 import SmallPrint from '../components/small-print.jsx';
 import SectionHeading from '../components/section-heading.jsx';
+import CurrencyDrop from '../components/currency-dropdown.jsx';
```
alicoding: And here

components/currency-dropdown.jsx
```diff
@@ -0,0 +1,18 @@
...
+
+module.exports = React.createClass({
+  render: function() {
+    var currencies = this.props.currencies;
+    return (
+      <select onChange={this.props.onChange} className="currency-dropdown" value={this.props.currency}>
+        {Object.keys(currencies).map(function(currency, i) {
+          return (
+            <option value={currency} key={i}>
+              {currency + " " + currencies[currency].symbol}
```
alicoding: I don't think we need a symbol beside the currency code here.

alicoding: We either going to just show currency code alone or showing something like this:

`Canadian Dollar (CAD)`

alicoding:
There is a bug when you pick a `NaN` and go back to reselect a valid number it doesn't seem to update the amount when trying to submit again. Here is the STR:

1. Set presets in the query string to `presets=1,2,3,a`
2. Select `NaN`
3. Fill in the form and try to donate
4. You should see an error which is expected.
5. Go back to step one of the donation flow and choose a valid amount
6. Try to donate again and it should allow you this time, but it is not.

components/currency-dropdown.jsx
ScottDowne: We also display symbol beside the other amount box. Is that ok there? Is it only the dropdown that is the issue?

ScottDowne:
Updated. r?

alicoding:
Looking good now. Just left you a comment there

mixins/form.jsx
```diff
@@ -272,7 +309,15 @@ module.exports = {
...
     var props = {};
     fields.forEach(function(name) {
-      props = assign(props, self.state[name].state.values);
+      var prop = self.state[name].state.values;
+      // Currently some fields expose their values on the form, and not themselves.
+      // So we need to check for props in both places until all field componenets are updated.
+      if (!prop) {
+        prop = self.state.props[name].values;
+      }
+      if (prop) {
```
alicoding: do you really need this check now? it looks like you are going to have prop no matter what?

ScottDowne:
@alicoding Updated

alicoding:
r+

PR #378, opened 2015-09-08
removing jquery for thank you page

ScottDowne:


ScottDowne:
This is reducing load times by removing dependencies, starting with Jquery and parsley for the thank you page. I couldn't remove the scripts themselves yet as I wanted to do an audit to make sure no other pages used them. For now, I focused just on thank you page.

This also gets us a step closer to supporting ie8.

I also use the form components I made for donations for the changes in the signup, which I know we wanted to do.

So three wins here. R?

components/email-input.jsx
```diff
@@ -39,14 +39,27 @@ var EmailInput = React.createClass({
...
+  renderHint: function() {
     var hintClassIconName = "fa fa-question-circle hint";
     var hintClassName = "hint-msg small";
+    var info = this.props.info;
     if (this.state.showHint) {
       hintClassIconName += " on";
     } else {
       hintClassName += " hidden";
     }
+    if (info) {
```
ScottDowne: Email component can be used in multiple forms, so making the hint generic and optional is probably ideal.

components/email-input.jsx
```diff
@@ -58,10 +71,7 @@ var EmailInput = React.createClass({
             <div className="field-container">
               <i className="fa fa-envelope field-icon"></i>
               <input type="email" className={inputClassName} name="email" value={this.state.values.email} onChange={this.onEmailChange} placeholder={this.getIntlMessage('email')}/>
-              <i onClick={this.hintClicked} className={hintClassIconName}></i>
-              <div className={hintClassName}>
-                <FormattedHTMLMessage message={ this.getIntlMessage("email_info") } />
-              </div>
+              {this.renderHint()}
```
ScottDowne: This is a new pattern I am trying for adding conditions in react markup. I kinda like it more than just making a whole new component because the `this` param is now the same in the conditional render, and no longer needs things passed to it.

components/signup.jsx
```diff
@@ -1,29 +1,11 @@
...
-      $email.val(donator_email).attr('value',donator_email);
-    }
-
-    $('#signup').parsley().subscribe("parsley:form:validate", function (formInstance) {
-      if (!formInstance.isValid()) {
-        formInstance.submitEvent.preventDefault();
-      }
-    });
-  },
+  mixins: [IntlMixin, require('../mixins/form.jsx')],
```
ScottDowne: This whole file is updated to use the email and privacy components, along with a submit button component.

components/submit-button.jsx
```diff
@@ -45,11 +24,7 @@ var DonateButton = React.createClass({
       <div className="row">
         <div className="full">
           <button onClick={this.onClick} type="submit" className="btn large-label-size" id="donate-btn">
-            <ButtonContent
-              amount={this.props.amount}
-              submitting={this.props.submitting}
-              currency={this.props.currency}
-            />
+            {this.renderButton()}
```
ScottDowne: We want to display "submitting..." with a spinner if submitting state is true, so again, I create a render function with my new attempt at accepting conditionals inside react markup.

I also boiled down this component to be generic. It submits to where ever, and displays a spinner or `this.props.children`

components/signup.jsx
```diff
@@ -1,29 +1,11 @@
 import React from 'react';
-import {IntlMixin, FormattedHTMLMessage} from 'react-intl';
+import IntlMixin from 'react-intl';
+import Email from '../components/email-input.jsx';
```
alicoding: EmailInput?

mixins/form.jsx
```diff
@@ -96,7 +95,7 @@ module.exports = {
     this.updateHeight();
   },
   submit: function(action, props, callback) {
-    props.localeCode = this.state.localeCode || "US";
+    props.locale = this.props.locales[0];
```
ScottDowne: We now get this from client.jsx. Feels more consistent to refer to locales in one way all over and define it as high level as we can.

alicoding: This is going to give you locale like this `en-US` vs `US`

pages/index.jsx
```diff
@@ -29,6 +23,12 @@ var Index = React.createClass({
           <div id="my-app" dangerouslySetInnerHTML={{__html: this.props.markup}}></div>
           <script src="https://cdn.polyfill.io/v1/polyfill.min.js?features=Intl.~locale.fr,Intl.~locale.pt-BR,Intl.~locale.de,Intl.~locale.id,Intl.~locale.es,Intl.~locale.en-US"></script>
           <script src="/main.js"></script>
+          <script src="https://js.stripe.com/v1/"></script>
+          <script src="/js/stripe.js"></script>
+          <script src="/js/stripe-checkout.js"></script>
+          <script src="/js/ga.js"></script>
+          <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
+          <script src="/js/parsley.min.js"></script>
```
ScottDowne: zzzzzzz boring change.

pages/sequential.jsx
```diff
@@ -9,7 +9,7 @@ import NavigationButton from '../components/navigation-button.jsx';
 import NavigationContainer from '../components/navigation-container.jsx';
 import NextButton from '../components/next-button.jsx';
 import Page from '../components/navigation-page.jsx';
-import DonateButton from '../components/donate-button.jsx';
+import SubmitButton from '../components/submit-button.jsx';
```
ScottDowne: generic submit button.

pages/sequential.jsx
```diff
@@ -21,8 +21,35 @@ import Address from '../components/address-input.jsx';
 import Email from '../components/email-input.jsx';
 import PrivacyPolicy from '../components/privacy-policy-input.jsx';
 
-var SingleForm = React.createClass({
-  mixins: [require('react-intl').IntlMixin, require('../mixins/form.jsx')],
+import {FormattedMessage, IntlMixin, FormattedNumber} from 'react-intl';
+
+var DonateButton = React.createClass({
```
ScottDowne: This is just the display for the submit button. This display makes it a donate button.

mixins/form.jsx
```diff
@@ -281,6 +280,28 @@ module.exports = {
...
+    if (result.error) {
+      this.setState({
+        errors: {
+          other: {
+            message: this.getIntlMessage('try_again_later')
+          }
+        }
+      });
+    } else {
+      window.location = '/' + this.props.locales[0] + '/share';
```
alicoding: I would rather us use react router API for consistent sake?

routes/index.js
```diff
@@ -4,20 +4,29 @@ var paypal = require('./paypal');
 
 var routes = {
   'signup': function(request, reply) {
-    var url = process.env.SINGUP;
+    var url = process.env.SIGNUP;
```
ScottDowne: Typo, was totally broken.

mixins/form.jsx
ScottDowne: I always forget about that thing. I can update.

mixins/form.jsx
ScottDowne: Isn't en-US what we want? At this point we're still fairly generic. I think "US" was just a mistake from the paypal locale, which, we should solve in a totally different way.

ScottDowne: To add more context.

I'm just passing along the locale for anyone on the other side (just our code) to handle with how it wants to.

This means it's up to us to filter out "en-US" to "US" where it makes sense.

alicoding: For URL en-US is what you want, but parsing it from locale to what PayPal expect might not be a good idea unless we have list of known one from PayPal and make a dictionary to lookup for that

components/signup.jsx
ScottDowne: I like Email. It's clear in these files that it's going to be an input.

mixins/form.jsx
ScottDowne: Yeah, I think we need something like that, and only do it for paypal.

Yes, I need it for url changes, paypal is the fringe case that only happens in one spot.

routes/index.js
alicoding: O_o

components/email-input.jsx
alicoding: Is `info` optional here?

components/email-input.jsx
alicoding: It kinda make sense for the same context, so I agree it is nice in this case.

components/signup.jsx
alicoding: I like Email too, but right now it seems like we lost all the `<input>` at the bottom there which makes all these `input` that you are adding not obvious until you look at the top here? I don't mind just the `Email` but I just want us to stick with something that will make it clear for other people to look and understand right away what's going on here, so if you think it is already very clear then leave it as is.

components/signup.jsx
alicoding: :+1:

components/submit-button.jsx
alicoding: I like it.

mixins/form.jsx
```diff
@@ -132,13 +133,8 @@ module.exports = {
...
 
     var params = '?payment=Stripe&str_amount=' + amount + '&str_currency=' + currency + '&str_id=' +transactionId + '&str_frequency=' +donationFrequency;
-    var thankYouURL = '/thank-you/' + params;
 
-    if (window.location.assign) {
-      window.location.assign(thankYouURL);
-    } else {
-      window.location = thankYouURL;
-    }
+    this.transitionTo('/thank-you/' + params);
```
alicoding: can you add `locale` here too?

pages/thank-you.jsx
```diff
@@ -32,7 +32,7 @@ var ThankYou = React.createClass({
                 </h2>
               </div>
             </div>
-            {this.props.params && ['de', 'en-US'].indexOf(locale) !== -1 ? <Signup language={locale}/> : <Social language={locale}/> }
+            {this.props.params && ['de', 'en-US'].indexOf(locale) !== -1 ? <Signup locales={this.props.locales}/> : <Social language={locale}/> }
```
alicoding: Any reason for this change?

components/email-input.jsx
ScottDowne: Yeah, singup doesn't have an info button.

It's the little "?" next to the email box in sequential, but I didn't see the need for it to always be there.

That said, if we want, we can create an info box for signup too, but would need new copy for the info. For now I'm fine with the way it is on prod.

mixins/form.jsx
ScottDowne: I can add it tomorrow.

pages/thank-you.jsx
ScottDowne: Yeah, being consistent between pages and components.

this.props.locale as an array is what we have inside all pages passed in from client.jsx. The form mixin mostly works with pages, so it assumes this locales prop exists.

If we want the form mixin to work with components and not just pages, we need the locale to have the same signature.

Honestly, we could also make it so client.jsx passes in language as a single locale code and not be an array, but I figured locales being an array was needed? Whatever we decide on, it needs to be this.props.localething in all places so the mixin can use it too.

ScottDowne: We could also define this.props.language along with this.props.locales, but the above felt like the easiest fix.

alicoding: Well the reason I asked because in social we use it differently here?

ScottDowne: Yeah, if and when social becomes an issue and needs to include the form mixin, we can update, but if not, it doesn't really bother me as is.

ScottDowne:
@alicoding Updated the locale passed into share now.

alicoding:
r+

alicoding:
you need to rebase now :P

PR #379, opened 2015-09-09
romanian translation

Jobava:
I am submitting a preliminary Romanian translation.

The Pontoon workflow is not so good at the moment, since sequential editing is not permitted. That means it's very tedious to edit a string, since once an edit is done you have to go hunting for the next string instead of that being automatically selected.

I expect to make future corrections from within Pontoon, however.

Please let me know if there are any issues with this edit. I merely copied the en-US folder, renamed it 'ro' and edited the files inside there.

Jobava


alicoding:
Hey @Jobava ! thank you for your PR! This is awesome.

PR #382, opened 2015-09-09
use visibility to stop tab nonsense on not active pages

ScottDowne:


ScottDowne:
R?

Turns out visibility has the functionality we need for stopping tabs, but can also be set as a transition. All the transition does is delay it, which is all we need of it.

PR #384, opened 2015-09-09
removed jquery and parsley

ScottDowne:


ScottDowne:
@alicoding thoughts?

components/donate-button.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import {FormattedMessage, IntlMixin, FormattedNumber} from 'react-intl';
+
+module.exports = React.createClass({
```
ScottDowne: Shared by sequential and simple paypal.

Not sure about the name. It's actually not much of a button, just kinda goes into a button.

components/simple-paypal.jsx
```diff
@@ -1,93 +1,30 @@
...
-          }
-          /> : this.props.defaultMessage}
-      </button>
-    );
-  }
-});
+import AmountButtons from '../components/amount-buttons.jsx';
+import Frequency from '../components/donation-frequency.jsx';
+import SubmitButton from '../components/submit-button.jsx';
+import DonateButton from '../components/donate-button.jsx';
```
ScottDowne: This whole form is updated to use the form mixing now too.

components/simple-paypal.jsx
```diff
@@ -1,93 +1,30 @@
...
-  componentDidMount: function() {
-    // ***********************************************
-    // Update Donate button to make it show the selected donation amount
-    // ***********************************************
-    var $theForm = $('#primary-form');
-
-    // check if a pre-selected amount is passed via the URL
-    if (window.location.hash.match(/#amount-\d+?/)) {
+  mixins: [IntlMixin, require('../mixins/form.jsx')],
+  simplePaypal(validate, props) {
```
ScottDowne: I kept this out of the mixin because it's only used in this file, and it was highly specific to the form data defined later in the file.

components/donate-button.jsx
alicoding: Can you add this as well

```
propTypes: {
    amount: React.PropTypes.number
  }
```

alicoding: wait ignore me

pages/index.jsx
```diff
@@ -27,8 +27,6 @@ var Index = React.createClass({
           <script src="/js/stripe.js"></script>
           <script src="/js/stripe-checkout.js"></script>
           <script src="/js/ga.js"></script>
-          <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.1.1/jquery.min.js"></script>
```
alicoding: :dancer:

components/donate-button.jsx
alicoding: actually yeah I think adding it doesn't hurt here since it is optional proptype anyway. I will let you decide here then.

PR #386, opened 2015-09-10
validate min amount for amout buttons.

ScottDowne:


ScottDowne:
r?

components/amount-buttons.jsx
```diff
@@ -70,7 +70,8 @@ var AmountButtons = React.createClass({
     }
     return {
       userInputting: userInputting,
-      valid: true
+      valid: true,
+      errorMessage: ""
```
ScottDowne: We now display multiple messages depending on the error, so we handle which error it is in the state.

components/amount-buttons.jsx
```diff
@@ -97,14 +98,40 @@ var AmountButtons = React.createClass({
   },
   validate: function() {
     var valid = false;
+    var errorMessage = "";
     if (this.props.amount) {
-      valid = true;
+      if (parseInt(this.props.amount, 10) < parseInt(this.props.currency.minAmount, 10)) {
```
ScottDowne: Check the donation amount against the min amount.

components/amount-buttons.jsx
```diff
@@ -97,14 +98,40 @@ var AmountButtons = React.createClass({
...
+              maximumFractionDigits={2}
+              value={this.props.currency.minAmount}
+              style="currency"
+              currency={this.props.currency.code || "usd"}
+            />
+          }
+        />
+      );
+    }
+    return this.state.errorMessage;
```
ScottDowne: Render the FormattedMessage is needed, or just return the error. The error could also be "" which means don't display anything or the default error.

components/simple-paypal.jsx
```diff
@@ -51,7 +51,12 @@ var simplePaypal = React.createClass({
             </div>
             <div className="row">
               <div className="full">
-                <AmountButtons currencySymbol={this.props.currency.symbol} currency={this.props.currency.code} onChange={this.onAmountChange} amount={amount} presets={this.props.presets} name="amount"/>
+                <AmountButtons name="amount"
+                  currency={this.props.currency}
```
ScottDowne: We're now passing in the currency as an object with symbol, code, and mine amount instead of passing them in as three props.

PR #388, opened 2015-09-10
validation for cc info

ScottDowne:


ScottDowne:
More simple validations where we can.

This restricts month exp, year exp, cc number and cvc to only be numeric keys, and also flags errors when you hit next for month and year values that are obviously wrong. being year is less than 15 and month is not between 1-12.

alicoding:
r+

PR #401, opened 2015-09-11
ensure stripe single donation and paypal single and monthly has curre…

ScottDowne:
…ncy changes, still not sure about stripe recurring.


ScottDowne:
r?

PR #402, opened 2015-09-11
allow frequency to be updated via query

ScottDowne:
Sooo, ?frequency=single/monthly query string.

This also means frequency is stored as a prop in the mixin, like amount, and not in the component, like in the other fields.

Again, eventually all fields need to be updated like this.

### DIFF
--- a/components/amount-buttons.jsx
+++ b/components/amount-buttons.jsx
@@ -83,7 +83,9 @@ var AmountButtons = React.createClass({
       userInputting: userInputting,
       valid: true
     });
-    this.props.onChange(this.props.name, this, amount);
+    this.props.onChange(this.props.name, this, {
+      amount: amount
+    });
   },
   otherRadioChange: function() {
     if (!this.state.userInputting) {
--- a/components/donation-frequency.jsx
+++ b/components/donation-frequency.jsx
@@ -2,44 +2,45 @@ import React from 'react';
 
 module.exports = React.createClass({
   mixins: [require('react-intl').IntlMixin],
-  getInitialState: function() {
-    return {
-      values: {
-        recurring: 0,
-        description: this.getIntlMessage("mozilla_donation")
-      }
-    };
+  componentDidMount: function() {
+    this.updateFrequency(this.props.value);
   },
   onChange: function(e) {
-    var recurring = e.currentTarget.value;
-    var description = this.state.description;
-    if (recurring) {
+    var value = "single";
+    if (e.currentTarget.value === "1") {
+      value = "monthly";
+    }
+    this.updateFrequency(value);
+  },
+  updateFrequency: function(value) {
+    var frequency = value;
+    var description = this.getIntlMessage("mozilla_donation");
+    if (frequency === "monthly") {
       description = this.getIntlMessage("mozilla_monthly_donation");
     }
-    this.setState({
-      values: {
-        frequency: recurring,
-        description: description
-      }
+    this.props.onChange(this.props.name, this, {
+      frequency: frequency,
+      description: description
     });
-    this.props.onChange(this.props.name, this);
   },
   validate: function() {
     return true;
   },
-  componentDidMount: function() {
-    this.props.onChange(this.props.name, this);
-  },
   render: function() {
+    var frequency = this.props.value;
     return (
       <div>
         <div className="row" id="recurring-option-row">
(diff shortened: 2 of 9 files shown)


components/amount-buttons.jsx
```diff
@@ -83,7 +83,9 @@ var AmountButtons = React.createClass({
       userInputting: userInputting,
       valid: true
     });
-    this.props.onChange(this.props.name, this, amount);
+    this.props.onChange(this.props.name, this, {
```
ScottDowne: The onChange now uses a generic function, because frequency has two props.

This means amount needs to pass in amount as a prop, because it *could* have multiple values.

components/donation-frequency.jsx
```diff
@@ -2,44 +2,45 @@ import React from 'react';
...
-    var recurring = e.currentTarget.value;
-    var description = this.state.description;
-    if (recurring) {
+    var value = "single";
+    if (e.currentTarget.value === "1") {
+      value = "monthly";
+    }
+    this.updateFrequency(value);
+  },
+  updateFrequency: function(value) {
```
ScottDowne: Creating a handy function `updateFrequency` so I can easily use it in the component did mount and the onchange.

components/donation-frequency.jsx
```diff
@@ -2,44 +2,45 @@ import React from 'react';
...
-    this.props.onChange(this.props.name, this);
   },
   validate: function() {
     return true;
   },
-  componentDidMount: function() {
-    this.props.onChange(this.props.name, this);
-  },
   render: function() {
+    var frequency = this.props.value;
```
ScottDowne: The value as to which is selected is now a static prop passed into here.

mixins/form.jsx
```diff
@@ -60,16 +66,15 @@ module.exports = {
     this.setState(newState);
     this.updateHeight();
   },
-  onAmountChange: function(name, value, amount) {
+  updateFormField: function(name, value, values) {
```
ScottDowne: This is now a generic function.

`updateFormField` is for props stored on the form mixin. `onchange` is a callback that does the similar thing, but used for handling props still stored in the components.

mixins/form.jsx
```diff
@@ -81,14 +86,10 @@ module.exports = {
       currency: currency.code,
       presets: presets.join(",")
     });
+    var newProps = this.state.props;
+    newProps.amount.values.amount = "";
     this.setState({
-      props: {
```
ScottDowne: Make sure we don't blow away other props, like frequency, when the currency changes. now set state on *just* amount to clear amount when the currency changes.

mixins/form.jsx
```diff
@@ -306,9 +307,13 @@ module.exports = {
     var self = this;
     var props = {};
     fields.forEach(function(name) {
-      var prop = self.state[name].state.values;
+      var state = self.state[name].state;
+      var prop;
       // Currently some fields expose their values on the form, and not themselves.
       // So we need to check for props in both places until all field componenets are updated.
+      if (state) {
```
ScottDowne: State may or may not be defined on a component.

example, amount component has a state object, and frequency does not. Neither of them have a values object though.

routes/index.js
```diff
@@ -31,7 +31,7 @@ var routes = {
   },
   'stripe': function(request, reply) {
     var transaction = request.payload || {};
-    if (transaction.recurring === 0) {
+    if (transaction.frequency === 'single') {
```
ScottDowne: Having a 0/1 values stored was hard to read. We also wanted the query string to be readable too. So I figure just always keep the same values here.

mixins/form.jsx
```diff
@@ -6,10 +6,11 @@ module.exports = {
   mixins: [Navigation],
   getInitialState: function() {
     var amount = "";
+    var frequency = "single";
     if (this.props.queryString) {
       amount = this.props.queryString.amount || "";
+      frequency = this.props.queryString.frequency || frequency;
```
alicoding: maybe do a check on that queryString.frequency if it is equal to `single` or `monthly` in case someone put crap in there.

alicoding:
r+

mixins/form.jsx
ScottDowne: Yeah, I can make this better.

PR #404, opened 2015-09-11
don't fail if we turn off optimizely

ScottDowne:
Set optimizely to no, and run npm start, without this patch, it'll fail.


alicoding:
r+

PR #405, opened 2015-09-12
Fix for IE11 missing feature "Promise"

alicoding:
IE11 and below doesn't support `Promise`, so here is the fix with Polyfill.io :)


alicoding:
@ScottDowne R?

PR #406, opened 2015-09-12
Fix IE10

alicoding:
IE10 doesn't support navigator.language, but we can default to
navigator.browserLanguage which is enough to get anyone to donate to us
based on their language setting of the browser


alicoding:
@ScottDowne R?

alicoding:
I have added more to that language list to make sure we can and will not break in case navigator.browserLanguage is undefined :)

PR #407, opened 2015-09-13
reduce complexity for values stored in frequency

ScottDowne:


components/donation-frequency.jsx
```diff
@@ -6,14 +6,9 @@ module.exports = React.createClass({
     this.updateFrequency(this.props.value);
   },
   onChange: function(e) {
-    var value = "single";
-    if (e.currentTarget.value === "1") {
-      value = "monthly";
-    }
-    this.updateFrequency(value);
```
ScottDowne: All the above logic is turning 1/0 into monthly/single

Why not just store the desired values in the inputs.

PR #410, opened 2015-09-14
clean up user inputting logic on amount buttons.

ScottDowne:
Small patch to reduce logic in amount buttons a bit.

### DIFF
--- a/components/amount-buttons.jsx
+++ b/components/amount-buttons.jsx
@@ -140,8 +140,8 @@ var AmountButtons = React.createClass({
     var amount = this.props.amount;
     var presets = this.props.presets;
     var preset = presets.indexOf(amount);
-    var userInputting = this.state.userInputting || (amount && preset < 0);
-    if (userInputting) {
+    var otherChecked = this.state.userInputting || (amount && preset < 0);
+    if (otherChecked) {
       otherAmount = amount;
       amount = "";
     }
@@ -165,7 +165,7 @@ var AmountButtons = React.createClass({
             onChange={this.onChange}/>
           <AmountOtherButton amount={otherAmount}
             currencySymbol={currency.symbol}
-            checked={userInputting}
+            checked={otherChecked}
             onRadioChange={this.otherRadioChange}
             onInputChange={this.otherInputChange}
             placeholder={this.getIntlMessage('other_amount')}
--- a/components/amount-buttons.jsx
+++ b/components/amount-buttons.jsx
@@ -31,6 +31,11 @@ var AmountOtherButton = React.createClass({
   onInputClick: function() {
     document.querySelector("#amount-other").click();
   },
+  onRadioChange: function() {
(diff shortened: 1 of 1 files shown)


components/amount-buttons.jsx
```diff
@@ -61,15 +66,10 @@ var AmountOtherButton = React.createClass({
 var AmountButtons = React.createClass({
   mixins: [require('react-intl').IntlMixin],
   getInitialState: function() {
-    var amount = this.props.amount;
-    var presets = this.props.presets;
-    var preset = presets.indexOf(amount);
-    var userInputting = false;
-    if (preset < 0 && amount) {
-      userInputting = true;
-    }
```
ScottDowne: We don't need to do any of the above in getInitialState. If we do it in render instead, we can get the same results.

It also cleans up the use of userInputting. How could user be inputting during getInitialState? Should always default to false.

components/amount-buttons.jsx
```diff
@@ -142,7 +140,7 @@ var AmountButtons = React.createClass({
     var amount = this.props.amount;
     var presets = this.props.presets;
     var preset = presets.indexOf(amount);
-    var userInputting = this.state.userInputting;
+    var userInputting = this.state.userInputting || (amount && preset < 0);
```
ScottDowne: This is all we need to replace the code removed in getInitialState.

components/amount-buttons.jsx
```diff
@@ -142,8 +140,8 @@ var AmountButtons = React.createClass({
     var amount = this.props.amount;
     var presets = this.props.presets;
     var preset = presets.indexOf(amount);
-    var userInputting = this.state.userInputting;
-    if (userInputting) {
+    var otherChecked = this.state.userInputting || (amount && preset < 0);
```
ScottDowne: This is all we need to add to get the same results as the code removed in getInitialState.

components/amount-buttons.jsx
```diff
@@ -31,6 +31,11 @@ var AmountOtherButton = React.createClass({
   onInputClick: function() {
     document.querySelector("#amount-other").click();
   },
+  onRadioChange: function() {
+    if (!this.props.checked) {
+      this.props.onRadioChange();
```
ScottDowne: We only want to fire onchanges if we go from not checked to cheked.

components/amount-buttons.jsx
```diff
@@ -88,9 +88,7 @@ var AmountButtons = React.createClass({
     });
   },
   otherRadioChange: function() {
-    if (!this.state.userInputting) {
-      this.setAmount("", true);
-    }
+    this.setAmount("", true);
```
ScottDowne: If we ensure onRadioChange only fires if it goes from not checked to checked, we can remove the if statement above. Better off to push that logic into the other buttons itself, because it's only needed in the other buttons.

PR #416, opened 2015-09-14
update preset on frequency

ScottDowne:
This updates presets when frequency changes.

It also makes it so changes to currency does not update the query string, so a refresh returns to the initial state.

### DIFF
--- a/data/currencies.js
+++ b/data/currencies.js
@@ -3,132 +3,198 @@ module.exports = {
     code: 'usd',
     minAmount: '2',
     symbol: '$',
-    presets: ['20', '10', '5', '3']
+    presets: {
+      single: ['20', '10', '5', '3'],
+      monthly: ['10', '5', '3', '2']
+    }
   },
   'aud': {
     code: 'aud',
     minAmount: '2',
     symbol: '$',
-    presets: ['20', '10', '5', '3']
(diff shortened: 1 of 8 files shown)


ScottDowne:
r?

data/currencies.js
```diff
@@ -3,132 +3,198 @@ module.exports = {
...
+      monthly: ['250', '200', '150', '100']
+    }
   },
   'thb': {
     code: 'thb',
     minAmount: '70',
     symbol: '฿',
-    presets: ['400', '200', '100', '70']
+    presets: {
+      single: ['3000', '2000', '1000', '500'],
```
alicoding: This is definitely way too high LOL

alicoding: but let's leave it like this and we do a research about this later.

alicoding:
r+

data/currencies.js
ScottDowne: It is based on unicef, which maybe is too high for us?

The min amount to lowest donation difference compared to our other currencies, might be the highest difference.

alicoding: UNICEF is a way bigger organization than us and their donors definitely not an average web users, so I don't think they will work for us here in my opinion.

PR #417, opened 2015-09-14
Update ENV and image to use local instead

alicoding:


webpack.config.js
```diff
@@ -9,6 +9,10 @@ var paths = require('./scripts/paths.js');
 var englishStrings = require('./locales/en-US.json');
 var currencies = require('./data/currencies.js');
 var url = require('url');
+var Habitat = require('habitat');
```
jbuck: Do this need the changes in this file?

alicoding: apparently we need this for `npm install` because of line 76 where we need to do `url.resolve(process.env.APPLICATION_URI...) and it is undefined.

webpack.config.js
```diff
@@ -30,14 +34,14 @@ module.exports = {
...
       'process.env': {
-        APPLICATION_URI: JSON.stringify(process.env.APPLICATION_URI),
-        STRIPE_PUBLIC_KEY: JSON.stringify(process.env.STRIPE_PUBLIC_KEY),
-        COINBASE_ENDPOINT: JSON.stringify(process.env.COINBASE_ENDPOINT),
-        OPTIMIZELY_ID: JSON.stringify(process.env.OPTIMIZELY_ID),
-        OPTIMIZELY_ACTIVE: JSON.stringify(process.env.OPTIMIZELY_ACTIVE),
-        FULL_SUBDOMAIN_FOR_COOKIE: JSON.stringify(process.env.FULL_SUBDOMAIN_FOR_COOKIE),
-        PAYPAL_EMAIL: JSON.stringify(process.env.PAYPAL_EMAIL),
-        PAYPAL_ENDPOINT: JSON.stringify(process.env.PAYPAL_ENDPOINT)
+        APPLICATION_URI: JSON.stringify(env.get('APPLICATION_URI')),
```
jbuck: Why are you stringifying a string? it's not already a string?

webpack.config.js
ScottDowne: Yeah he walked me through this. Right now we oonly use habitat inside server.js.

During a npm install, we run webpack without server.js, thus the envs are undefined.

PR #420, opened 2015-09-15
busted stripe errors

ScottDowne:
I just missed some state scope while doing a recent patch, and broke the error handling.

STR:
1. test a declined card: 4000000000000002
2. fill in the rest of the form with valid test data.

Expected: should swing you to page 2 with an error about the card being declined
Actual: just spins, and eventually throws a console error with something being undefined.


ScottDowne:
R?

PR #421, opened 2015-09-16
fixing paypal locales

ScottDowne:
Fixing the locale for paypal's login screen.

Few things to note:

1. once logged in, it uses that account's settings locale, and our current test account is a US account. We just control the login screen locale.

2. It also for some reason caches the last used locale, and ignore the one we set. So to test each of these, you need to test them in new private tabs.

### DIFF
--- a/intl-config.json
+++ b/intl-config.json
@@ -1,7 +1,7 @@
 {
-	"supportedLocales": ["en-US", "de", "fr", "pt-BR", "es", "id"],
-	"dest": "locales",
-	"src": "locales",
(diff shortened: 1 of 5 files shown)


ScottDowne:
R?

intl-config.json
```diff
@@ -1,5 +1,12 @@
 {
 	"supportedLocales": ["en-US", "de", "fr", "pt-BR", "es", "id"],
 	"dest": "locales",
-	"src": "locales"
+	"src": "locales",
+  "paypalLocales": {
```
alicoding: indentation is weird

PR #423, opened 2015-09-16
Fix eslint to lint jsx as well

alicoding:
Currently it doesn't lint any JSX, so this patch should now enable that.


alicoding:
@ScottDowne R?

PR #424, opened 2015-09-16
adding params into paypal redirect to reduce to one api path, and to …

ScottDowne:
…redirect to locale thank you page

### DIFF
--- a/routes/index.js
+++ b/routes/index.js
@@ -80,64 +80,58 @@ var routes = {
   },
   'paypal': function(request, reply) {
     var transaction = request.payload || {};
-    if (transaction.frequency !== 'monthly') {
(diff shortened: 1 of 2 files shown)


ScottDowne:
R?

ScottDowne:
This does one fundamental change, adding params to the paypal redirect. One for frequency and one for locale.

This ends up making two things better.

1. We no longer have to api routes for redirect, one being single donation, and one being monthly. We can now just pass the frequency and have one path. This reduces bugs caused by fixing one route and forgetting to fix the other, and it means I didn't need to copy and paste similar code between the two routes.
2. It allows the thank you page after the redirect to have a locale.

PR #425, opened 2015-09-16
Add eslint-loader

### LINKED ISSUE #422: Add support for ESLint in webpack
When developing, webpack currently doesn't lint the project. It's possible to not notice linting issues until changes are pushed and Travis flags them. We should eslint to the watch task so that warnings are logged out on the console on file updates and can be fixed immediately.

alanmoo:
ESlint warnings/errors will show in console when developing

Fixes #422


alicoding:
r+

PR #429, opened 2015-09-21
playing with dynamic google font load

ScottDowne:


MikkCZ:
IMO this should be included into [New Fundraising Platform Launch](https://github.com/mozilla/donate.mozilla.org/milestones/New%20Fundraising%20Platform%20Launch) milestone, as blocks Czech (at least) to be displayed correctly.

ScottDowne:
A couple external issues and projects threw us off those milestones. So we're behind.

ScottDowne:
Updated. @alicoding r?

I removed it from the list of supported locales, as I wasn't sure what needs to happen for new languages to be added.

This isn't a question for just cs, now that we have so many more translated, we could probably turn on a bunch.

PR #432, opened 2015-09-22
improve email error

### LINKED ISSUE #414: Validate for email address on CC personal details form
Currently the forms submits even if the email suplied isn't valid:

screenshot 2015-09-14 14 00 14

ScottDowne:
What this is doing is differentiating between invalid emails and empty inputs.

I got the translated string from the petition, along with the English version so I can be pretty sure they match.

Fixes #414


ScottDowne:
R?

PR #433, opened 2015-09-23
Fix IE9

alicoding:
I didn't fix placeholder in this patch, but at least this will make the page load.


alicoding:
@ScottDowne R?

PR #434, opened 2015-09-23
Update copy for snippet

alicoding:


ScottDowne:
Typo "If the didn't, we couldn't do all this good  stuff"

The "If the didn't" is off.

PR #435, opened 2015-09-23
Fix issue #304 - Browser support

### LINKED ISSUE #304: With the changes to client-side i10n, what browser coverage do we have?
From the tech-review. Can we confirm the answer to this question:

>* With the changes to client-side i10n, what browser coverage do we have?
  * (Old browser support?)
    * it only detects locale if a locale is NOT in the url (I think...)

alicoding:
Closes #304


alicoding:
@ScottDowne R?

PR #438, opened 2015-09-23
enable different plans for different currencies

ScottDowne:


ScottDowne:
R?

routes/index.js
```diff
@@ -54,14 +54,15 @@ var routes = {
         amount: transaction.amount,
         stripeToken: transaction.stripeToken,
         email: transaction.email,
+        currency: transaction.currency,
         metadata: {
-          firstname: transaction.firstName,
-          lastname: transaction.lastName,
+          firstname: transaction.first,
```
jbuck: I almost wonder if we shouldn't just be creating a Stripe customer object first...

ScottDowne: It's required, and I do create the customer inside the function I send this data to.

https://github.com/ScottDowne/donate.mozilla.org/blob/batmans-got-plans/routes/stripe.js#L22

ScottDowne: Sounds like though the order of things is not obvious or optimal maybe?

ScottDowne:
This had a merge conflict, but it's fixed now.

R? @jbuck or @alicoding whoever gets here first?

PR #444, opened 2015-09-24
Make value names a bit more consistent where we use them so we can be…

ScottDowne:
… sure they are cleared on input, and handle the correct type vs rawType from stripe errors.

### DIFF
--- a/mixins/form.jsx
+++ b/mixins/form.jsx
@@ -25,10 +25,10 @@ module.exports = {
       errors: {
         creditCardInfo: {
           page: 0,
-          number: "",
+          cardNumber: "",
           cvc: "",
-          monthExp: "",
-          yearExp: ""
+          expMonth: "",
+          expYear: ""
         },
         address: {
           page: 0,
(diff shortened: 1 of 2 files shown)


mixins/form.jsx
```diff
@@ -264,13 +265,13 @@ module.exports = {
       exp_year: submitProps.expYear
     }, function(status, response) {
       if (response.error) {
-        error(response.error);
+        error(response.error.code, response.error.type);
```
ScottDowne: This is the main fix. createToken errors return error.type, and stripe.charges.create return error.rawType.

Both are the values we want.

mixins/form.jsx
```diff
@@ -264,13 +265,13 @@ module.exports = {
...
       } else {
         submitProps.cardNumber = "";
         submitProps.stripeToken = response.id;
         submit("/api/stripe", submitProps, function(result) {
           if (result.error) {
-            error(result.error);
+            error(result.error.code, result.error.rawType);
```
ScottDowne: See above comment.

mixins/form.jsx
```diff
@@ -25,10 +25,10 @@ module.exports = {
       errors: {
         creditCardInfo: {
           page: 0,
-          number: "",
+          cardNumber: "",
           cvc: "",
-          monthExp: "",
-          yearExp: ""
+          expMonth: "",
+          expYear: ""
```
ScottDowne: Just making these the same as other similar values we use in other places.

ScottDowne:
R?

alicoding:
r+ from me

PR #445, opened 2015-09-24
minify js for build not watch

ScottDowne:


ScottDowne:
R? this is a P2 but was a quick win.

alicoding:
r+

PR #450, opened 2015-09-25
Fix issue #443 - Localize site title

### LINKED ISSUE #443: Localize HTML title tag
For example, on the DE version of the form, we should localize the current title tag text `Give to Mozilla Today`

screenshot 2015-09-24 14 05 23

alicoding:
Closes #443


alicoding:
@ScottDowne R?

PR #458, opened 2015-09-29
Add Karma tests

alicoding:


alicoding:
@ScottDowne I don't want this patch to get any bigger, so this should be the initial tests and the setup so that we can add "more" tests :dancer:

components/name-input.jsx
```diff
@@ -14,13 +18,13 @@ var NameInput = React.createClass({
   },
   validate: function() {
     var valid = true;
-    if (!this.state.values.first) {
+    if (this.state.values.first.trim(' ').length === 0) {
```
ScottDowne: trim doesn't accept any props, it by default works with whitespace.

tests/react/index.jsx
```diff
@@ -0,0 +1,9 @@
+/*eslint no-unreachable: 0*/
+// Make console.warn throw
+var warn = console.warn;
+console.warn = function (warning) {
+  throw new Error(warning);
+  warn.apply(console, arguments);
+};
+
+require('./inputs.test.jsx');
```
ScottDowne: I think a better name might be input.tests.jsx

webpack.config.js
```diff
@@ -51,6 +51,7 @@ module.exports = {
     new webpack.ProvidePlugin({
       'fetch': 'imports?this=>global!exports?global.fetch!whatwg-fetch'
     }),
+    new webpack.ContextReplacementPlugin(/buffer/, require('buffer')),
```
ScottDowne: What's this doing?

package.json
```diff
@@ -62,9 +78,11 @@
     "react-ga": "^1.0.14",
     "react-intl": "^1.2.0",
     "react-router": "^0.13.3",
+    "react-test-context": "0.0.2",
     "request": "^2.60.0",
     "scooter": "^2.0.1",
     "selenium-webdriver": "^2.46.1",
+    "should": "^7.1.0",
```
ScottDowne: These seem like a lot. Are you sure all of these are needed?

components/name-input.jsx
ScottDowne: Also, can probably first check if it's false so we can skip checking trim if we know it's false.

It also ensure we can never call trim on a non string value just in case (although right now we are defaulting).

Something like:

```
if (!this.state.values.first || !this.state.values.first.trim()) {}
```

webpack.config.js
alicoding: There is a bug in Buffer module inside webpack, so what's this doing is making sure `buffer` version that we specified in package.json is used.

package.json
alicoding: Yep

webpack.config.js
ScottDowne: So this has nothing to do with the new tests, this is just in general broken?

karma.conf.js
```diff
@@ -0,0 +1,35 @@
...
+    reporters: ['mocha'],
+    coverageReporter: {
+      type: 'lcov',
+      dir: 'www/coverage/'
+    },
+    webpack: {
+      devtool: 'inline-source-map',
+      module: {
+        loaders: webpackConfig.module.loaders,
+        postLoaders: process.env.KARMA_ENV === 'coverage' ? [{
```
ScottDowne: can we put a sample.env for this?

ScottDowne: I see what you're doing here, you set it when you call it. I guess this is fine.

tests/react/name-input.tests.jsx
```diff
@@ -0,0 +1,70 @@
...
+    it('should not throw when passing all props', function () {
+      should.doesNotThrow(() => {
+        var NameInput = stubContext(Name, IntlContext);
+        TestUtils.renderIntoDocument(<NameInput name='ali' onChange={function(){}} />);
+      });
+    });
+    it('.validate() should return false when first name and last name field are empty', function () {
+      should.doesNotThrow(() => {
+        var NameInput = stubContext(Name, IntlContext);
+        var el = TestUtils.renderIntoDocument(<NameInput name='ali' onChange={function(){}} />);
```
ScottDowne: name is the name of the input, and not the input value. Is use on all donation inputs.

Example: https://github.com/mozilla/donate.mozilla.org/blob/master/pages/single-form.jsx#L32

It is totally confusing on a name input, thought. So I'm open to new names for this prop.

ScottDowne: However, I would have used firstName="" and lastName="" for this.

karma.conf.js
alicoding: yeah we don't want to have default for this, so only set when we run that script :+1:

webpack.config.js
alicoding: yeah this is broken if you do npm install for webpack with latest version, so this is just workaround for getting webpack with this dependency to work for now until they patch this upstream.

tests/react/name-input.tests.jsx
alicoding: I just want to pass in some random string there, but if you want to change the prop name `name` maybe we can change it in another patch or something?

alicoding:
@ScottDowne test now should run for travis

tests/react/name-input.tests.jsx
ScottDowne: just change it to name="name-input"

ScottDowne:
Seems Travis is running, and failing?

alicoding:
@ScottDowne travis is green!

Issue #454, opened 2015-09-29
Some currencies don't accept decimals


ScottDowne:
Just reading this: https://developer.paypal.com/docs/classic/api/currency_codes/

Turns out some currencies don't allow decimals in their amount, an example is jpy.

We should handle this.

ScottDowne:
PayPal gives a generic error on their site.

ScottDowne:
Here is Stripes zero decimal currencies. https://support.stripe.com/questions/which-zero-decimal-currencies-does-stripe-support

This is also a bit more complex as Stripe accepts decimal currencies in their lowest common currency unit, so, we convert 1$ to 100cents for decimal currencies. We need to make sure we don't do this calculation to yen, as an example.

ScottDowne:
This is a P1 as yen for stripe isn't just throwing an error, but it's actually converting it by 100 as if it was cents... inflating the donation by 100.

ScottDowne:
Fixed: https://github.com/mozilla/donate.mozilla.org/commit/74a99775872b8a7b23d4036481246c2dc6a5b1f2

PR #459, opened 2015-09-29
ensure we handle currencies with zero decimals for both stripe and pa…

ScottDowne:
…ypal

### DIFF
--- a/data/currencies.js
+++ b/data/currencies.js
@@ -93,6 +93,7 @@ module.exports = {
     code: 'huf',
     minAmount: '490',
     symbol: 'Ft',
+    zeroDecimal: "paypal",
     presets: {
       single: ['5000', '3500', '2000', '1000'],
       monthly: ['2000', '1500', '1350', '1000']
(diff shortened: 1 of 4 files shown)


ScottDowne:
Fixes #454 

What to do to trigger previously busted paths, and what to expect.

for all zero decimal currencies, we trunc the cents. Just not worth it anyway. it's fractions of a penny.

This is enough for paypal.

Stripe however, also needs to trunc, and needs to ensure we don't do the convert from dollars to cents, it's already in cents.

Try making a stripe donation with 1000 yen, and the thank you page should have 1000 in the query string. Can also look in the stripe logs if you setup your own test stripe account and plug those keys into the env.

Now try another Stripe donation with 1000.1, should also end up as just 1000 in the query (drop the .1)

Dollars should still work the way they did before, so 10$ should be 1000 in the thank you page query.

Paypal should just drop the decimal, and not do any modifying of the unit (dollars to cents) so try 1000.1 and it should end up as 1000 in the paypal page.

PR #463, opened 2015-09-30
simple paypal doing amount modifier and moving amount modifier module…

ScottDowne:
… into scripts dir

### DIFF
--- a/scripts/amount-modifier.js
+++ b/scripts/amount-modifier.js
@@ -2,17 +2,20 @@ var currencies = require('../data/currencies.js');
 
 var amountModifier = {
   modify: function(amount, service, currencyCode) {
+    if (!amount) {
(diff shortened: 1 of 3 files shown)


ScottDowne:
R?

scripts/amount-modifier.js
```diff
@@ -2,17 +2,20 @@ var currencies = require('../data/currencies.js');
 
 var amountModifier = {
   modify: function(amount, service, currencyCode) {
+    if (!amount) {
+      return "";
```
alicoding: return `amount` wouldn't be better?

ScottDowne: amount might be false, 0, NaN, undefined, etc.

Just being sure here.

PR #479, opened 2015-10-01
Add address-input-tests

alicoding:


alicoding:
Review using this @ScottDowne https://github.com/mozilla/donate.mozilla.org/pull/479?w=1

mixins/input.jsx
```diff
@@ -3,7 +3,7 @@ module.exports = {
     var self = this;
     var valid = true;
     fields.forEach(function(field) {
-      if (self.props.error[field] || !self.state.values[field]) {
+      if (self.props.error[field] || !self.state.values[field].trim()) {
```
ScottDowne: It's good practice before using trim to first make sure it's not false.

undefined.trim is a common console error.

So should look something like:

```
if (self.props.error[field] || !self.state.values[field] || !self.state.values[field].trim()) {
```

ScottDowne: You could also load up a variable for this, and default it to an empty string:

```
var str = self.state.values[field] || "";
if (self.props.error[field] || !str.trim()) {
```

PR #487, opened 2015-10-02
trying signup using basket and exact-target

ScottDowne:


mixins/form.jsx
```diff
@@ -137,7 +137,10 @@ module.exports = {
   },
   submit: function(action, props, callback) {
     props.locale = this.props.locales[0];
-    props.currency = this.state.currency.code;
+    var currency = this.state.currency;
+    if (currency) {
+      props.currency = currency.code;
+    }
```
ScottDowne: This was just a straight up bug that was blocking signups.

state.currency only exists for donations, singup was throwing a console error.

mixins/form.jsx
```diff
@@ -350,15 +353,15 @@ module.exports = {
     }
   },
   signup: function(validate, props) {
-    this.setState({
-      submitting: true
-    });
```
ScottDowne: This was creating a waiting spinner even if we know it was going to fail. I moved it down to ensure we only start the spinner if basic validation passes.

ScottDowne:
R?

ScottDowne:
I tried this with ten minute email, and it worked.

So likely it was the fact I was already signed up that was causing this to succeed but do nothing.

routes/index.js
```diff
@@ -7,22 +7,24 @@ var routes = {
...
-      url = url + '-' + transaction.locale;
-    }
-    httpRequest({
+    httpRequest.post({
       url: url,
-      method: 'POST',
       form: {
-        'opt_in': '1',
+        format: 'html',
+        lang: transaction.locale,
```
adamlofting: This is fine, so long as we're only showing the sign-up thank you page for `en` and `de` donors. We haven't setup mailing list for other locales.

adamlofting:
R+

routes/index.js
ScottDowne: Yup, we handle that here: https://github.com/mozilla/donate.mozilla.org/blob/master/pages/thank-you.jsx#L35

Issue #465, opened 2015-09-30
Add user data validation for query parameters, headers, and payload body


adamlofting:
Action from #82 

- Applied to all user controlled data
 - [x] Form data
 - [x] URL parameters
 - [x] Hidden fields
 - [x] Cookie data
 - [x] HTTP Headers
 - [x] Essentially anything in the HTTP request
- Define the type of data (String, Number, etc)
- Define the minimum & maximum length of the data
- Must be done client-side and server-side

PR #488, opened 2015-10-02
Fix issue #465 - validate data

alicoding:


alicoding:
@ScottDowne R?

components/client.jsx
```diff
@@ -33,11 +33,14 @@ Router.run(routes, Router.HistoryLocation, function (Handler, state) {
       frequency = "monthly";
     }
   }
-  var currency = currencies[currencyCode];
-  presets = presets.split(",");
 
+  var currency = Object.keys(currencies).indexOf(currencyCode) === -1 ? currencies['usd'] : currencies[currencyCode];
```
ScottDowne: I find this pretty cryptic.

What do you think about something like:

```
var currency = currencies[currencyCode];
if (!currenct) {
  currenct = currencies['usd'];
}
```

components/client.jsx
```diff
@@ -33,11 +33,14 @@ Router.run(routes, Router.HistoryLocation, function (Handler, state) {
...
   // We didn't get valid presets from the query string,
   // so default to the currency and frequency preset.
+  if(/^[\d,]+$/.test(presets)) {
```
ScottDowne: `?presets=asdf` creates 4 NaN values.

PR #498, opened 2015-10-05
optimize stripe params

ScottDowne:
This is changing it so we only pass what we need to the stripe thank you page.


ScottDowne:
R?

PR #500, opened 2015-10-06
hapi and joi for signup

ScottDowne:

PR #501, opened 2015-10-06
Hapi joi paypal

ScottDowne:


ScottDowne:
This builds on the #500 

So I think I'll close #500.

alicoding:
This is looks good to me, so r+

PR #522, opened 2015-10-08
Add validation for query amount

alicoding:


alicoding:
@ScottDowne R?

ScottDowne:
Looks good, add tests in a new PR, because changes need to be made to make it testable.

PR #528, opened 2015-10-09
Test for amount-modifier

alicoding:


alicoding:
@ScottDowne R?

tests/lib/amount-modifier.tests.js
```diff
@@ -0,0 +1,43 @@
...
+    amountModifier.modify(false, 'paypal', 'usd').should.equal('0');
+  });
+  it("'abcd', 'PayPal', 'usd' 'abcd'", function() {
+    amountModifier.modify('abcd', 'paypal', 'usd').should.equal('abcd');
+  });
+  it("'10.10', 'PayPal', 'huf' '10'", function() {
+    amountModifier.modify('10.10', 'paypal', 'huf').should.equal('10');
+  });
+  it("10.10, 'PayPal', 'huf' '10'", function() {
+    amountModifier.modify(10.10, 'paypal', 'huf').should.equal('10');
```
ScottDowne: One thing I think we can also add is similar to this test.

This test is kinda testing two things. It's testing that modifyer works with a 0 cent currency.

But it's also testing that huf is a 0 cent currency.

twd is also a zero cent currency, so maybe we can add a test just like the above but change huf for twd and one ofr jpy?

alicoding:
@ScottDowne added another test.

PR #533, opened 2015-10-10
first pass at reducing down shared.less

ScottDowne:


ScottDowne:
Couple goals here:
- remove more code than add
- easier to manage less files
- no visual changes (if I fix a bug or two, there might be visual changes from that)

@alicoding I'll be doing this in blocks, and pushing it up to this pull request as I finish up blocks and keep it in a stable state, so whatever state this pull request is in, it's review able.

If you decide to land, my next block just becomes a new pull request.

PR #535, opened 2015-10-12
Add app.json to allow review app on heroku

alicoding:


alicoding:
@jbuck R?

jbuck:
@alicoding I don't think this configuration will work. You'll need to have the heroku application name exposed into APPLICATION_URI somehow.

PR #536, opened 2015-10-13
update click area for navigation buttons

ScottDowne:


ScottDowne:
Box shadow doesn't create a clickable area, border does.

Border is included in the width, box shadow is not.

So if we change it to border, and update the height and width to be the total size (previous size plus shadow) we get the same result.

PR #543, opened 2015-10-14
Require === and !== (eqeqeq)

alicoding:
It is considered good practice to use the type-safe equality operators
=== and !== instead of their regular counterparts == and !=.


alicoding:
@ScottDowne R?

PR #544, opened 2015-10-14
dissallow return in else (no-else-return)

alicoding:
If an if block contains a return statement, the else block becomes
unnecessary. Its contents can be placed outside of the block.


alicoding:
@ScottDowne R?

PR #551, opened 2015-10-15
update donation_components.md

imjared:
in return for the hundreds of hours of work you've put into this repo, please accept some very thoughtful and useful work from me.


alicoding:
Thank you very much for the PR :P I think newxt time we will put more time on the readme :P

PR #555, opened 2015-10-15
Make share url configurable

alicoding:


alicoding:
@ScottDowne R?

ScottDowne:
I'm not convinced, either we always point to prod, and it always "works"

Or we link to application uri and it's busted locally.

alicoding:
@ScottDowne I would rather have this busted locally than not able to test on Facebook debugger or share popup.

ScottDowne:
@alicoding I didn't try this, but wouldn't this be resolved now that prod exists?

jbuck:
@ScottDowne I think it should be configurable so we can test share changes locally and on staging, rather than testing on production. r+

ScottDowne:
@jbuck I would agree with you, but that's not what this is doing.

Try to share locally now, nothing happens. facebook !== local

ScottDowne:
Which might be fine.

jbuck:
With ngrok it would!

PR #565, opened 2015-10-16
add ability to flip frequency

ScottDowne:


ScottDowne:
?test=flip-frequency to turn it on.

Only one test can be run via query string at a time, but optimizely can turn on multiples at a time.

ScottDowne:
R?

PR #585, opened 2015-10-19
Optimize webpack config file a bit

alicoding:
I felt like webpack config is too big and just abstract the stuff related to creating files to its own file, so that we can easily change without making webpack anymore painful to look at.


alicoding:
@ScottDowne R?

PR #587, opened 2015-10-19
improve button style for test

ScottDowne:


ScottDowne:
Previously, the little line around the border was a shadow, and the line around the border was a border.

This was problematic because it mean the border line around the border didn't have a gradient.

Now it does, it's another inner element with a transparent border that's 1 pixel inside the parent.

ScottDowne:
r?

PR #588, opened 2015-10-19
fixing height transition

ScottDowne:


ScottDowne:
Using ?test=new-test-flow there is a noticeable little height transition at the bottom of the next button.

This is due to our form previously having a static height.

I removed this with a absolute/relative trick.

ScottDowne:
r?

PR #589, opened 2015-10-19
merge simple paypal and paypal donate into one page

ScottDowne:


ScottDowne:
Basically, we had a time where simple paypal was generating multiple types of pages (local+currency)

So we had a component for it.

I don't see the value in having this anymore.

ScottDowne:
r?

PR #603, opened 2015-10-20
new donation flow for thank you and share pages

ScottDowne:


ScottDowne:
use ?test=new-flow-test on share and thank-you pages.

components/new-signup.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
```
alicoding: Where are we using this file again?

PR #605, opened 2015-10-21
qa on share links

ScottDowne:


ScottDowne:
R?

components/social-test.jsx
```diff
@@ -14,17 +14,17 @@ module.exports = React.createClass({
           </h2>
           <div className="row">
             <div className="half" id="facebook">
+              <div className="social-button"></div>
```
ScottDowne: These buttons were not clickable because of the pseudo element.

I had to make the a tag and the thing with the border inside the same html scope so they could be siblings and then the a could be on top, so it can be clicked.

PR #608, opened 2015-10-21
handle stripe for countries without an obvious province, and display …

ScottDowne:
…an error to try again later if our joi ever fails again.


ScottDowne:
R?

Stripe is failing right now for countries without known provinces.

It also doesn't handle errors from joi very well, just spins, so I also fixed that.

ScottDowne:
@jbuck feel free to take this before @alicoding

jbuck:
@ScottDowne r+

PR #612, opened 2015-10-22
fix strip checkout

ScottDowne:


ScottDowne:
R?

server.js
```diff
@@ -102,6 +102,35 @@ server.route([
     }
   }, {
     method: 'POST',
+    path: '/api/stripe-checkout',
```
ScottDowne: Stripe checkout is never going to have address, so instead of making address bits optional in the stripe route, I created a new one that only has what we need.

So if we try to use stripe checkout WITH address, it'll fail.

mixins/form.jsx
```diff
@@ -314,15 +314,21 @@ module.exports = {
...
     var handler = StripeCheckout.configure({
       // Need to get this from .env
       key: process.env.STRIPE_PUBLIC_KEY,
       image: '',
       token: function(response) {
         // Where is this things error? Maybe it's not called at all for an error case.
-        submitProps.cardNumber = "";
-        submitProps.stripeToken = response.id;
-        submit("/api/stripe", submitProps, success);
+        submit("/api/stripe-checkout", {
```
ScottDowne: Let's just pass in exactly what we need to stripe checkout.

mixins/form.jsx
```diff
@@ -314,15 +314,21 @@ module.exports = {
...
+          frequency: submitProps.frequency,
```
alicoding: This will always be `single` right?

alicoding: Otherwise this will fail if you pass `monthly`

ScottDowne: Hm, monthly should be fine, and it was fine when I tested it.

ScottDowne: Why would it not be fine or fail?

alicoding: It will go to the second condition in 'stripe' server api where address is expected to be passed

ScottDowne: You mean the metadata object is going to be empty? https://github.com/mozilla/donate.mozilla.org/blob/master/routes/index.js#L68-L76

Items in the metadata object are totally optional, if I recall correctly.

Is it not working for you? Worked for me, but I'll try again when I get a chance.

alicoding: hmm I didn't test the code, but I thought sending undefined to Stripe is that okay?

ScottDowne: I think so.

It's also not a new thing I'm changing in this patch.

Before we broke this, it was using the same path and passing in undefined.

PR #619, opened 2015-10-23
Add script to create Stripe subscription plans

jbuck:


jbuck:
@alicoding r?

ops/stripe_plans.js
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env node
+
+var async = require("async");
+var currency_data = require("../data/currencies");
+var currencies = Object.keys(currency_data);
```
ScottDowne: Maybe make the top two lines one and remove the currency_data var?

Is there something significant to the name "currency_data" or could it be "currenctData"?

ScottDowne: I notice you using the underscore in filenames again. I don't really have an issue with it, other than we don't follow that naming convention. If you like the underscore then we can keep it.

ScottDowne:
Looks good to me.

PR #624, opened 2015-10-26
Dynamic locales

### LINKED ISSUE #491: turn on a bunch of languages
We have a bunch of languages in varying degrees of translated, some I would bet are as complete as our main 5. I'm not sure how to confirm the completeness, though, so I'm just guessing :P

We have a list in our repo to say "we allow these x languages" anything else redirects to en-us, which is pretty obnoxious. Especially if you're a translator working on one of those unsupported but finished languages.

I say we turn these all on, even if not 100%.

I think this is safe as long as we don't link to an unfinished language in the snippet.

@alicoding thoughts?

alicoding:
Closes #491

### DIFF
--- a/intl-config.js
+++ b/intl-config.js
@@ -1,5 +1,7 @@
-{
-  "supportedLocales": ["en-US", "de", "fr", "pt-BR", "es", "id"],
+require('habitat').load();
+
+module.exports = {
+  "supportedLocales": process.env.SUPPORTED_LOCALES || ["en-US", "de", "fr", "pt-BR", "es", "id"],
   "dest": "locales",
   "src": "locales",
   "paypalLocales": {
@@ -9,4 +11,4 @@
     "pt-BR": "BR",
     "id": "id_ID"
   }
-}
+};
--- a/intl-config.js
+++ b/intl-config.js
@@ -1,7 +1,7 @@
 require('habitat').load();
 
 module.exports = {
-  "supportedLocales": process.env.SUPPORTED_LOCALES || ["en-US", "de", "fr", "pt-BR", "es", "id"],
+  "supportedLocales": process.env.SUPPORTED_LOCALES || "*",
   "dest": "locales",
   "src": "locales",
   "paypalLocales": {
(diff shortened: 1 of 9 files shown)


alicoding:
This will allow us to enable all locales on staging server by just setting the `env` to `*` that way we will always have newly added locale enabled for testing.

alicoding:
@ScottDowne you can test this by adding `SUPPORTED_LOCALES` to your `.env` and set it to `*` and then try run the script by running `npm run build:l10njson`

scripts/properties2json.js
```diff
@@ -26,10 +45,13 @@ function getJSON(locale) {
 }
 
 function processMessageFiles(locales) {
-  return Promise.all(locales.map(getJSON));
+  if (config.supportedLocales !== "*") {
```
ScottDowne: It looks like this is deciding to use locales or not after getting them.

Flow looks like:

1. `getListLocales` returns locales.
2. check if we want to use those locales or not.

I think we can optimize this by reversing this order. First check if we need all locales or not, then call `getListLocales` by seeing if `config.supportedLocales !== "*"` is true.

intl-config.js
```diff
@@ -1,5 +1,7 @@
-{
-  "supportedLocales": ["en-US", "de", "fr", "pt-BR", "es", "id"],
+require('habitat').load();
+
+module.exports = {
+  "supportedLocales": process.env.SUPPORTED_LOCALES || ["en-US", "de", "fr", "pt-BR", "es", "id"],
```
ScottDowne: Can we default this to "*" after the `||`?

alicoding: I don't want to default this to `*` since our production is not going to have this enabled by default.

alicoding: Unless we want to make sure they always have to set this for production...

ScottDowne: I think it's reasonable to have this be a config for prod.

If you think about it, the only place where it's not going to be "*" is prod. So prod really is the only place that needs this to be configurable. Making me feel like this config is for prod.

It also makes it easier to update the list.

ScottDowne:
Looks good, and thanks for considering my nits.

Once this is green, you can land.

PR #625, opened 2015-10-26
add test for signup in third step

ScottDowne:


ScottDowne:
R?

The test string to test this is "?test=signup-test"

This should enable another checkbox in the third step. It should also ensure your thank you page is a share page if you signup in the third page.

server.js
```diff
@@ -88,8 +89,9 @@ server.route([
           id: Joi.string(),
           frequency: Joi.string().valid("monthly", "one-time"),
           currency: Joi.string().min(3).max(3),
-          quantity: Joi.string(),
-          amount: Joi.number()
+          quantity: Joi.number(),
```
ScottDowne: quantity is actually passes as a number, and not a string. Fixing. This is busted on staging.

routes/signup.js
```diff
@@ -0,0 +1,18 @@
+var httpRequest = require('request');
+
+module.exports = function(transaction, callback) {
```
ScottDowne: moved signup into a module because it's called in multiple spots now, because we do this inside stripe now too.

components/checkbox.jsx
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import { FormattedHTMLMessage, IntlMixin } from 'react-intl';
+
+var Checkbox = React.createClass({
+  mixins: [IntlMixin],
+  getInitialState: function() {
+    var values = {};
+    values[this.props.type] = false;
+    return {
```
alicoding:
```
return {
  values: {
                 [this.props.type]: false
  },
  valid: true
}
```

components/checkbox.jsx
```diff
@@ -0,0 +1,95 @@
...
+    this.props.onChange(this.props.name, this);
+  },
+  onChange: function(e) {
+    this.setState({
+      valid: true
+    });
+    var values = {};
+    values[this.props.type] = e.currentTarget.checked;
+    this.setState({
+      values: values
```
alicoding: same here...

PR #644, opened 2015-10-29
remove mixin input because it wasn't really being used anymore

ScottDowne:


ScottDowne:
R? @alicoding 

At one point this was used a bit more, but it could only be used for quite basic cases, and it ended up being removed from all but one component.

I just merged the small bits of it that was still being used into the only component that was still using it.

ScottDowne:
Also the only component that still uses it I think is the only component that still stores it's values internally.

Once we update creditCard to store it's values on the form mixin, I think all the bits would be removed.

PR #647, opened 2015-10-30
Fix issue #643 - Enable other English locales to see the signup

### LINKED ISSUE #643: Tweak L10N to capture other en locales (inc en-gb)
@alicoding - what are your thoughts on this? Do we just duplicate the strings, or can we set this automatically somewhere to re-use the en-us strings? That might be easier for maintenance.

--

As a start, what can we do to setup an en-gb locale?

This thank you page sends me straight to the social share, but actually anyone who speaks an en variant (including the :crown: 's English) is a valid candidate to join our mailing list.
https://donate.mofostaging.net/en-GB/thank-you/

Also, should we setup an en-gb locale so that we can default the currency to £GBP for people in the UK. And even en-ie to default to €EUR?

alicoding:
closes #643

### DIFF
--- a/pages/thank-you.jsx
+++ b/pages/thank-you.jsx
@@ -19,6 +19,10 @@ var ThankYou = React.createClass({
   },
   renderBaseline: function() {
     var locale = this.props.locales[0];
+    var signUpOrSocial = (<Social language={locale}/>);
+    if (this.props.params && /^(?=.*\ben\b)|^(?=.*\bde\b)/.test(locale)) {
+      signUpOrSocial = (<Signup country={this.props.country} email={this.props.email} locales={this.props.locales} />);
+    }
     return (
       <span className="base-line-thank-you">
         <Header>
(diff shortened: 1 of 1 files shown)


alicoding:
@ScottDowne R?

alicoding:
P.S. Don't review this patch until Monday @ScottDowne ! It's your birthday, so don't work! O_O

pages/thank-you.jsx
```diff
@@ -19,6 +19,10 @@ var ThankYou = React.createClass({
   },
   renderBaseline: function() {
     var locale = this.props.locales[0];
+    var signUpOrSocial = (<Social language={locale}/>);
+    if (this.props.params && /^(?=.*\ben\b)|^(?=.*\bde\b)/.test(locale)) {
```
alicoding: Basically anything that starts with `en*` and `de` will get the signup the rest default to social.

alicoding:
@ScottDowne can you review this?

pages/thank-you.jsx
ScottDowne: Could we use something like this? `'en-US'.indexOf("en") === 0`

ScottDowne: ^(en|de)(\b|$)

PR #649, opened 2015-10-30
More events

ScottDowne:


ScottDowne:
WIP

ScottDowne:
I think this is ready for review.

1. I'm removing all callbacks and turning them into events.
2. Because they are events, I had to store the data where the listeners existed, which meant removing all cases where we stored the events internally to the components.
3. Because it's stored externally, it no longer has a concept of a component wrapper, like creditCardInfo wrapping all credit card fields. Which meant I also had to split the creditCardInfo components into components for each field.
4. I pretty much forced the react input tests to be completely different, and they are not really testing much anymore. The Selenium tests cover quite a bit now, so I'm OK with this for the moment.

ScottDowne:
This is replaced by https://github.com/mozilla/donate.mozilla.org/pull/707

PR #652, opened 2015-10-31
Fix selenium test and remove :poop:

alicoding:


alicoding:
@ScottDowne R?

ScottDowne:
loooool

PR #653, opened 2015-10-31
default country or amount without a change causes a failed submit

ScottDowne:


ScottDowne:
If you try to use the form right now without changing the country (leave it as us) it'll fail on submit.

This is a bug that's exposed when I uplift the values from inside the components to into the form. It needs to be registered a bit better, for now, making this a change seems easiest. I'm working on a patch that fixes it in a more fundamental way, but I don't plan on working on it for a few days and wanted to get something quick.

PR #654, opened 2015-10-31
Simple selenium test runner

ScottDowne:


ScottDowne:
I added a simple test runner to run multiple selenium tests from multiple files.

I also added a selenium for the thank-you page, which is kinda busted right now...

@alicoding r?

alicoding:
https://www.npmjs.com/package/selenium-test-runner?

ScottDowne:
I couldn't figure it out. Also looked like it runs tests on only one url?

ScottDowne:
Looking at it a bit more, I think it's meant to work with the downloaded selenium and not the npm selenium-driver thingy we use.

tests/selenium/start.js
```diff
@@ -5,64 +5,16 @@ var driver = new webdriver.Builder().
   withCapabilities(webdriver.Capabilities.firefox()).
   build();
 
+var testRunner = require('./test-runner');
+var sequentialForm = require('./sequential-form-test.js')(driver, By);
```
alicoding: Why passing `By`? Why not just require it there since we don't even use it in this file?

ScottDowne: EH, just trying it with one less thing to remember to require in for each test file.

ScottDowne: Makes for a tighter test file.

ScottDowne: I also do all requiring in this file, currently.

PR #656, opened 2015-11-02
adding events for navigation and ensuring we properly remove events

ScottDowne:


ScottDowne:
In person R+ with @alicoding

PR #657, opened 2015-11-03
fixing a weird race condition

ScottDowne:


ScottDowne:
I noticed if you told the test runner to run three tests, it started to error out.

I tested this by running the thank-you test three times.

I then turned off the bit that turned off the server and driver, and noticed the tests was being run more than three times. The test runner was doing someting weird with the ordering. Firing functions and not removing them at the right moment, then turning off the server before calling these leaked tests, thus the error.

Ultimatly this is due to me trying to chain these things. I went back to a single call with a static array (and made it a bit better than the first implementation), seems to be fixed now.

tests/selenium/sequential-form-test.js
```diff
@@ -1,7 +1,6 @@
 module.exports = function(driver, By) {
   function test(noProvince, url) {
     var country = 'Canada';
-    driver.manage().timeouts().implicitlyWait(3000);
```
ScottDowne: We only need to do this once, at the start of running all tests, and not at the start of each test. Just some cleanup.

tests/selenium/sequential-form-test.js
```diff
@@ -46,13 +45,12 @@ module.exports = function(driver, By) {
       return driver.getCurrentUrl().then(function(url) {
         return url.indexOf('http://localhost:3000/de/thank-you/') === 0;
       });
-    }).then(function() {
```
ScottDowne: This doesn't actually need to be in a then.

Webdriver already links it's wait functions. We just need to be sure the done bit is inside a then (as the done is not a webdriver thing and thus does not wait.).

tests/selenium/start.js
```diff
@@ -5,16 +5,20 @@ var driver = new webdriver.Builder().
   withCapabilities(webdriver.Capabilities.firefox()).
   build();
 
+driver.manage().timeouts().implicitlyWait(3000);
```
ScottDowne: We do this once now. Right here.

tests/selenium/test-runner.js
```diff
@@ -1,18 +1,8 @@
-module.exports = function() {
-  var tests = [];
+module.exports = function(tests) {
```
ScottDowne: Got it down to six lines.

tests/selenium/thank-you-test.js
```diff
@@ -1,17 +1,16 @@
 module.exports = function(driver, By) {
-  function test(url) {
-    driver.manage().timeouts().implicitlyWait(3000);
```
ScottDowne: Removed too.

tests/selenium/thank-you-test.js
```diff
@@ -1,13 +1,9 @@
...
-    driver.get(url);
+  return function(done) {
+    driver.get('http://localhost:3000/en-US/thank-you/');
     driver.findElement(By.name('email')).sendKeys('test@email.com');
     driver.findElement(By.id('privacy-policy-checkbox')).click();
     driver.findElement(By.css('.submit-btn')).click();
-  }
-  return function(done) {
-    test('http://localhost:3000/en-US/thank-you/');
```
ScottDowne: Figured this wasn't really needed in this test file.

ScottDowne:
@alicoding no rush on this r?

PR #660, opened 2015-11-03
initial drop of a bunch of selenium tests, they pass mostly... but ha…

ScottDowne:
…ve odd failures. The driver crashes sometimes.


ScottDowne:
When you try to run them all, they have intermittent failures. Running one at a time, by commenting out the rest via start.js they tend to be more likely to pass.

I'm OK with this for now, and kinda not surprised that these tests have intermittent failures.

PR #663, opened 2015-11-03
ensure stripe passes and uses address info

ScottDowne:


ScottDowne:
@jbuck r?

This is address metadata for stripe customers.

ScottDowne:
I also did this for the one-page form along with a few other decent options after exploring the API more.

ScottDowne:
Found an issue with this.

ScottDowne:
Fixed up the issue and updated the test that landed after I started this.

ScottDowne:
@jbuck feel free to review this too.

tests/selenium/single-page-stripe-test.js
```diff
@@ -11,6 +11,11 @@ module.exports = function(driver, By, done) {
     });
   });
   driver.findElement(By.id('email')).sendKeys('send-donation@test.com');
+  driver.findElement(By.id('billing-name')).sendKeys('a');
```
ScottDowne: Just updating the test, which now uses a bunch of new fields so we need to fill them in.

server.js
```diff
@@ -111,7 +112,14 @@ server.route([
           amount: Joi.number().required(),
           frequency: Joi.string().min(6).max(7).required(),
           stripeToken: [Joi.string().required(), Joi.number().required()],
-          locale: Joi.string().min(2).max(12).required()
+          locale: Joi.string().min(2).max(12).required(),
```
ScottDowne: checkout expects a few more props available to it.

routes/stripe.js
```diff
@@ -8,17 +8,30 @@ var stripe = require('stripe')(stripeKeys.secretKey);
...
+        stripe.charges.create(charge, callback);
+      }
+    });
   },
   recurring: function(transaction, callback) {
     stripe.customers.create({
       email: transaction.email,
-      metadata: transaction.metadata
+      metadata: transaction.metadata,
+      source: transaction.stripeToken
```
ScottDowne: Moving source (the user's credit card number) from subscription to customer. Wasn't a huge reason, but had to go somewhere and something about capturing it sooner rather than on a second request felt marginally better.

routes/stripe.js
```diff
@@ -8,17 +8,30 @@ var stripe = require('stripe')(stripeKeys.secretKey);
 
 module.exports = {
   single: function(transaction, callback) {
-    var charge = {
-      amount: transaction.amount,
-      currency: transaction.currency,
-      card: transaction.stripeToken
-    };
-    stripe.charges.create(charge, callback);
+    stripe.customers.create({
```
ScottDowne: For single donations we also store a customer.

This means we can capture their address and thus have better fraud control.

routes/index.js
```diff
@@ -21,11 +21,24 @@ var routes = {
...
+      locale: transaction.locale
+    };
     if (transaction.frequency !== 'monthly') {
       stripe.single({
         amount: amount,
         currency: currency,
-        stripeToken: transaction.stripeToken
+        stripeToken: transaction.stripeToken,
+        email: transaction.email,
+        description: transaction.description,
```
ScottDowne: Capture description because we can.

routes/index.js
```diff
@@ -54,20 +67,10 @@ var routes = {
...
-          lastname: transaction.last,
-          country: transaction.country,
-          address: transaction.address,
-          city: transaction.city,
-          zip: transaction.code,
-          state: transaction.province,
-          locale: transaction.locale,
-          signup: transaction.signup
-        }
+        metadata: metadata
```
ScottDowne: metadata is the same for monthly and single, so let's cache it above.

routes/index.js
```diff
@@ -54,20 +67,10 @@ var routes = {
...
-        metadata: {
-          firstname: transaction.first,
-          lastname: transaction.last,
-          country: transaction.country,
-          address: transaction.address,
-          city: transaction.city,
-          zip: transaction.code,
-          state: transaction.province,
-          locale: transaction.locale,
-          signup: transaction.signup
```
ScottDowne: signup was removed because it actually had no business in here.

We use is in this scope, but don't need to pass it to stripe.

mixins/form.jsx
```diff
@@ -342,13 +347,23 @@ module.exports = {
     if (!valid) {
       return;
     }
+    var description = this.getIntlMessage("mozilla_donation");
+    var handlerDesc = "";
     submitProps = this.buildProps(props);
+    if (submitProps.frequency === "monthly") {
+      description = this.getIntlMessage("mozilla_monthly_donation");
+      handlerDesc = this.getIntlMessage("monthly");
```
ScottDowne: Display in the strpe widget that it's monthly, just as a handy reminder. Single just displays nothing, because I think it's safe to assume it's implied.

mixins/form.jsx
```diff
@@ -342,13 +347,23 @@ module.exports = {
...
+
     var locale = this.props.locales[0];
     var currency = this.state.currency && this.state.currency.code;
     var handler = StripeCheckout.configure({
       // Need to get this from .env
       key: process.env.STRIPE_PUBLIC_KEY,
       image: '',
+      zipCode: true,
+      billingAddress: true,
+      locale: locale,
```
ScottDowne: Moving locale into the config and not the open. This is fairly semantic, but I'm going for configure setsup the layout and context of the widget, and open fills in the layout and context with data.

mixins/form.jsx
```diff
@@ -342,13 +347,23 @@ module.exports = {
...
+    }
+
     var locale = this.props.locales[0];
     var currency = this.state.currency && this.state.currency.code;
     var handler = StripeCheckout.configure({
       // Need to get this from .env
       key: process.env.STRIPE_PUBLIC_KEY,
       image: '',
+      zipCode: true,
+      billingAddress: true,
```
ScottDowne: These say use zip and address in the widget.

mixins/form.jsx
```diff
@@ -293,10 +293,14 @@ module.exports = {
     if (!valid) {
       return;
     }
+    var description = this.getIntlMessage("mozilla_donation");
```
ScottDowne: We pass along description where we can.

routes/stripe.js
```diff
@@ -8,17 +8,30 @@ var stripe = require('stripe')(stripeKeys.secretKey);
...
+      email: transaction.email,
+      metadata: transaction.metadata,
+      source: transaction.stripeToken
+    }, function(err, customer) {
+      var charge = {};
+      if (err) {
+        callback(err);
```
alicoding: return too?

alicoding:
r+ with just one comment to fix.

PR #665, opened 2015-11-04
fix intermittent selenium failures caused by reusing driver

ScottDowne:


ScottDowne:
R?

As the title implies, reusing the driver was leaving context, or something.

Tests running on their own were fine, but running them all back to back caused failures.

By ditching it on each new test file, I stopped getting failures.

tests/selenium/sequential-paypal-test.js
```diff
@@ -1,26 +1,24 @@
...
-  return function(done) {
-    driver.get('http://localhost:3000/en-US/');
-    driver.findElement(By.id('amount-other-input')).sendKeys('10');
-    driver.findElement(By.css('.page-active .next-button')).click();
-    driver.findElement(By.css('.paypal-button .payment-submit-button')).click();
-    driver.wait(function() {
-      return driver.getCurrentUrl().then(function(url) {
-        return url.indexOf('https://www.sandbox.paypal.com/cgi-bin/webscr') === 0;
-      });
+module.exports = function(driver, By, done) {
```
ScottDowne: All these tests files got tabbed, and had their two functions merged.

ScottDowne: This is because we no longer store the context for driver and by in a separate scope from where done comes from.

Because they all come from the same scope, I can pass them all in one shot, thus only need one function.

tests/selenium/test-runner.js
```diff
@@ -1,8 +1,22 @@
+var server = require('../../server');
```
ScottDowne: test-runner now handles server, driver, and By...

It also handles calling next, and setting up and taking down driver and By.

ScottDowne: The handling of next is basically the same as before, cept now I plug into certain points and do driver related stuff.

tests/selenium/start.js
```diff
@@ -1,32 +1,16 @@
...
-driver.manage().timeouts().implicitlyWait(50000);
-
 var testRunner = require('./test-runner');
-var sequentialStripe = require('./sequential-stripe-test.js')(driver, By);
-var sequentialPaypal = require('./sequential-paypal-test.js')(driver, By);
-var singlePageStripe = require('./single-page-stripe-test.js')(driver, By);
-var singlePagePaypal = require('./single-page-paypal-test.js')(driver, By);
-var simplePaypal = require('./simple-paypal-test.js')(driver, By);
-var thankYou = require('./thank-you-test.js')(driver, By);
+var sequentialStripe = require('./sequential-stripe-test.js');
```
ScottDowne: Pretty small file now, just handles loading tests and firing the test runner.

PR #667, opened 2015-11-04
ensure we can do a monthly donation from simple paypal

ScottDowne:


ScottDowne:
The monthly donate paypal page looks different than the single payment.

Right now using /paypal-donate/ click monthly, and hit donate, you get the single payment page.

This is just because it was still checking against the previous API we used for this. (I don't think it ever worked in the new form, as the API it expected was what we used in BSD)

ScottDowne:
@alicoding r?

PR #669, opened 2015-11-04
remove all envs for bitcoin to keep only the envs that we actually ne…

ScottDowne:
…ed to be configurable


ScottDowne:
Scratch that, we decided to make this static to keep down the amount of envs we have.

R? @jbuck

jbuck:
@ScottDowne r+

PR #685, opened 2015-11-05
pass email and country to thank you page, and use email if it is avai…

ScottDowne:
…lable.


ScottDowne:
R? @alicoding or @jbuck

PR #693, opened 2015-11-05
display a generic required fields error message when required

ScottDowne:


ScottDowne:
R? @alicoding

alicoding:
can you rebase

ScottDowne:
Needs a bit more work than a rebase, API changed a bit, but should be hard to adapt.

Issue #691, opened 2015-11-05
Form is not showing up in Safari and Safari Mobile


valianttry:
This is what i see at donate.mofoprod.net on my iPhone 6s using Chrome:

![img_1242](https://cloud.githubusercontent.com/assets/9830371/10984779/cd7f1ca6-83d0-11e5-8a17-d497618ee942.PNG)

valianttry:
Note same sadness ensues when I us Chrome on iOS to see donate.mofostaging.net as well. :cry:

adamlofting:
Just checked, and I get the same on Safari on iPhone.

jbuck:
It won't work on Safari because our Intl polyfill is not being loaded

jbuck:
We're good now: https://www.dropbox.com/s/bj33cddpl9rx5ju/Screenshot%202015-11-06%2014.24.50.png?dl=0

PR #698, opened 2015-11-06
Update to polyfill-service@2.1.0 & default to using polyfill

jbuck:


server.js
```diff
@@ -180,11 +180,12 @@ server.route([
       var polyfills = PolyfillSet.fromQueryParam(features || 'default', flags);
       var params = {
         features: polyfills.get(),
-        minify: true
+        minify: true,
+        unknown: 'polyfill'
```
jbuck: This changes the behaviour of the polyfill service to default to using all of the polyfills when it cannot detect the user-agent correctly.

jbuck:
@ScottDowne r?

jbuck:
Fixes #691

ScottDowne:
R+

PR #699, opened 2015-11-06
fix amount-modifier to handle cents correctly

ScottDowne:


scripts/amount-modifier.js
```diff
@@ -1,19 +1,29 @@
...
 var amountModifier = {
+  convertToCents: function(amount) {
+    amount = amount || "0";
+    if (typeof amount !== 'string') {
+      amount += "";
+    }
+    var splitAmount = amount.split(".");
+    var dollar = splitAmount[0];
+    var cents = splitAmount[1] || "0";
+    return (parseFloat(dollar, 10) * 100) + parseFloat(cents, 10);
```
jbuck: You could use parseInt here, right?

mixins/form.jsx
```diff
@@ -400,7 +401,7 @@ module.exports = {
       description: handlerDesc,
       currency: currency,
       // Stripe wants cents.
-      amount: submitProps.amount * 100
+      amount: amountModifier.stripe(submitProps.amount, currency)
```
jbuck: Wouldn't you want to use modify here?

scripts/amount-modifier.js
```diff
@@ -1,19 +1,29 @@
 var currencies = require('../data/currencies.js');
 
 var amountModifier = {
+  convertToCents: function(amount) {
```
jbuck: I think this should be an internal API only, not something outside users can access

scripts/amount-modifier.js
```diff
@@ -1,19 +1,30 @@
...
     if (!amount) {
       return "0";
     }
-    var amt = parseFloat(amount, 10);
     var currency = currencies[currencyCode];
     var zeroDecimal = currency.zeroDecimal || "";
+    // These currencies don't use cents
     if (zeroDecimal.indexOf(service) > -1) {
-      return Math.trunc(amt) + "";
+      return Math.trunc(parseFloat(amount, 10)) + "";
```
jbuck: Lets change this to parseInt as well

PR #712, opened 2015-11-09
adding selenium test for sequential page with reduced address

ScottDowne:


ScottDowne:
This adds a new selenium test for sequential page with reduced address.

It also properly sets the name prop in places where it was failing.

This was because my selenium test for an AB test we trying to click an element not displayed because of name conflicts. By using the name prop on this, it clicks the same prop that is registered via the form.

ScottDowne:
r?

Issue #701, opened 2015-11-06
"Next" button shows little margin lines in iPhone 6s


valianttry:
![img_1246](https://cloud.githubusercontent.com/assets/9830371/11007507/31228640-847f-11e5-9e8a-687ee9551976.PNG)

ScottDowne:
@valianttry can you check this against the new flow?

donate.mofoprod.net/?test=new-flow-test

valianttry:
@ScottDowne I used the form at mozilla.org (live form) and still see the little lines:

![img_1326_png](https://cloud.githubusercontent.com/assets/9830371/11049958/03c1a0ee-86f7-11e5-8282-46808dc068b5.png)

ScottDowne:
I tested this on an ipad, and I saw it in the old flow, but it was gone in donate.mozilla.org/?test=new-flow-test

ScottDowne:
Fixed, should be on staging soon.

PR #717, opened 2015-11-10
fixing weird white lines in next button arrow when iphone or other we…

ScottDowne:
…ird zoom conditions

### DIFF
--- a/less/components/next-button.less
+++ b/less/components/next-button.less
@@ -11,13 +11,15 @@
   height: 46px;
   cursor: pointer;
   &::before {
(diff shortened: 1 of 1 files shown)


ScottDowne:
@jbuck r?

ScottDowne:
Fixes https://github.com/mozilla/donate.mozilla.org/issues/701

It was easy and I knew we had a fix because the new flow button didn't have this issue. So I just copied that css for both buttons.

jbuck:
@ScottDowne r+

PR #724, opened 2015-11-10
send locale as meta data, and nothing else, address and name instead …

ScottDowne:
…are already sent when making the charge, so let's no duplicate that.


jbuck:
@ScottDowne r+

PR #726, opened 2015-11-11
quick easy fix to give a bit more room to display amounts the fill th…

ScottDowne:
…e buttons

### DIFF
--- a/components/amount-buttons.jsx
+++ b/components/amount-buttons.jsx
@@ -10,8 +10,8 @@ var AmountButton = React.createClass({
 
     return (
       <div className="third">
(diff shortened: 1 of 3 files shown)


less/shared.less
```diff
@@ -273,7 +273,7 @@ input[name="donation_amount"] + label,
 input[name="payment-type"] + label {
   display: inline-block;
   width: 100%;
-  padding: 6px 12px;
+  padding: 6px 0;
```
jbuck: This also affects the placement of the $ symbol in the last box

jbuck:
r-

ScottDowne:
I cleaned up some of the css class naming so I could more specifically target the thing I wanted in a clear way.

ScottDowne:
@jbuck r?

jbuck:
@ScottDowne r+

PR #727, opened 2015-11-11
We broke the next button for new flow when we made a change to the re…

ScottDowne:
…gular button


ScottDowne:
To test this go to the ?test=new-flow-test page, and look at the button. It has some noise from the baseline page.

Just need to clear these css props for the test.

ScottDowne:
r?

jbuck:
@ScottDowne r+

PR #730, opened 2015-11-11
Fix dropdown weird position on Chrome

alicoding:
This annoys me a bit that in Chrome the currency dropdown looks really
weird


ScottDowne:
r+

PR #755, opened 2015-11-12
Use throng module to enable usage of node Clustering

### LINKED ISSUE #197: Look at enabling node cluster module usage
Heroku recommended looking into using the node cluster module. This basically lets you run multiple instances of a node app listening to the same port. This is especially required for PX dynos.

cadecairos:
Fixes #197

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "should": "7.1.1",
     "simple-html-precompiler": "1.0.1",
     "stripe": "4.0.0",
-    "throng": "^1.0.1",
+    "throng": "1.0.1",
     "webpack": "1.12.2",
     "whatwg-fetch": "0.10.0"
   },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -93,6 +93,7 @@
     "should": "7.1.1",
     "simple-html-precompiler": "1.0.1",
     "stripe": "4.0.0",
+    "throng": "^1.0.1",
```
jbuck: Can you use an exact version?

server.js
```diff
@@ -16,261 +16,246 @@ var polyfillio = require('polyfill-service');
 var PolyfillSet = require('./scripts/PolyfillSet.js');
```
jbuck: I think you'll need to move some/all of these variable declarations into the export function

jbuck:
@cadecairos r+

Issue #767, opened 2015-11-12
Presets not working


valianttry:
![donate_mozilla_org___spenden_sie_noch_heute_an_mozilla](https://cloud.githubusercontent.com/assets/9830371/11133850/6322f76a-894d-11e5-8982-1f6dc65724c2.png)

And on live US-EN 
![donate_mozilla_org___give_to_mozilla_today](https://cloud.githubusercontent.com/assets/9830371/11133890/a30f3aa0-894d-11e5-84b4-e9421a42794a.png)
form linked from mozilla.org:

ScottDowne:
Fixed on staging: https://donate.mofostaging.net/de/?presets=1,2,3,4

should be live in a few moments.

jbuck:
Fixed on production now too!

PR #768, opened 2015-11-12
ensure we default form mixin values before we setup listeners

ScottDowne:


jbuck:
@ScottDowne r+

jbuck:
Fix #767

PR #794, opened 2015-11-14
add a test for signup with prefilled country, it is optional, and mer…

ScottDowne:
…ged some test code into easier to maintain blocks because I was finding too much copy and pasted code in my patch.


ScottDowne:
r?

PR #798, opened 2015-11-14
fixing and updating some tests after API changes

ScottDowne:


ScottDowne:
Karma tests for input components are now useful again, and nothing was broken.

I updated name attribute on cc inputs to be the name passed to it and not something static. I then changed any references to the static name to be id instead which is probably a better thing to use to reference this.

ScottDowne:
@alicoding R? on Monday?

PR #800, opened 2015-11-15
required fields error message

ScottDowne:


ScottDowne:
It works well, but the css isn't great for the required error message on the third page.

alicoding:
tests are failing

ScottDowne:
@alicoding yup, which is why it's not in review :P

ScottDowne:
@alicoding hey sassy pants, mind doing this review now? Should be fixed :D

components/address-input.jsx
```diff
@@ -624,12 +624,15 @@ var Country = React.createClass({
     this.setState({
       valid: true
     });
+    // Changing the country should clear any errors on province.
+    form.error("province", "");
```
alicoding: What is empty string there?

alicoding:
r+

components/address-input.jsx
ScottDowne: Clears the error message.

ScottDowne: If province was an error, changing country should clear any errors on the country dropdown left over from the previous country's list of provinces, if any.

PR #802, opened 2015-11-15
Add transaction_id string to be translated

jbuck:


jbuck:
@alicoding I'm 99% sure this is fine, but just want to verify it's all good

PR #806, opened 2015-11-16
Hashes in file names

jbuck:
This patch allows us to put hashes in file names.


jbuck:
@alicoding lets pick this apart!

jbuck:
@cadecairos I'd love to get your input on this patch

ScottDowne:
I give this an R+ not only because the selenium tests pass, but because I agree with where it's going.

PR #813, opened 2015-11-17
Make a request to specific version of polyfill

alicoding:


jbuck:
Nice win @alicoding! r+

PR #814, opened 2015-11-17
adding a few more string from snippet and some other small changes fo…

ScottDowne:
…r snippet locales


ScottDowne:
@jbuck Small r?

I'm fine with not accepting signups for pt, fr, and es but if we can make it happen, it would probably be good.

data/locale-data.js
```diff
@@ -6,7 +6,8 @@ module.exports = {
     'es': 'eur',
     'en-US': 'usd',
     'en-CA': 'cad',
-    'en-GB': 'gbp'
+    'en-GB': 'gbp',
+    'it': 'eur'
```
ScottDowne: I added this when I thought we were doing IT but, having it here is harmless.

ScottDowne:
@alicoding r?

PR #820, opened 2015-11-17
Directory tiles for l10n

alicoding:


alicoding:
@ScottDowne if you are okay can you hit merge for me please!

PR #834, opened 2015-11-18
adding currency change and presets change to event system

ScottDowne:
This is a WIP

### DIFF
--- a/components/amount-buttons.jsx
+++ b/components/amount-buttons.jsx
@@ -20,7 +20,7 @@ var AmountButton = React.createClass({
       <div className="third">
         <input onChange={this.props.onChange} checked={checked} className="amount-radio" type="radio" name="donation_amount" value={this.props.value} id={"amount-" + this.props.value}/>
         <label htmlFor={"amount-" + this.props.value} className="amount-button large-label-size">
-          { this.props.currencyCode ?
+          { this.props.currencyCode && this.props.value ?
           <FormattedNumber
             minimumFractionDigits={0}
             value={this.props.value}
--- a/components/amount-buttons.jsx
+++ b/components/amount-buttons.jsx
@@ -16,17 +16,17 @@ var AmountButton = React.createClass({
     if (this.props.value === this.props.amount) {
       checked = true;
     }
-
     return (
       <div className="third">
         <input onChange={this.props.onChange} checked={checked} className="amount-radio" type="radio" name="donation_amount" value={this.props.value} id={"amount-" + this.props.value}/>
         <label htmlFor={"amount-" + this.props.value} className="amount-button large-label-size">
+          { this.props.currencyCode ?
           <FormattedNumber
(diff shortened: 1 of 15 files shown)


ScottDowne:
No longer a WIP, it's not a shiny new patch. @alicoding r?

components/amount-buttons.jsx
```diff
@@ -16,17 +16,17 @@ var AmountButton = React.createClass({
...
         <label htmlFor={"amount-" + this.props.value} className="amount-button large-label-size">
+          { this.props.currencyCode && this.props.value ?
           <FormattedNumber
             minimumFractionDigits={0}
             value={this.props.value}
             style="currency"
-            currency={this.props.currencyCode || "usd"}
-          />
+            currency={this.props.currencyCode}
+          /> : <span>&nbsp;</span> }
```
alicoding: why you need a whitespace?

components/amount-buttons.jsx
```diff
@@ -64,7 +64,12 @@ var AmountOtherButton = React.createClass({
             value={this.props.amount}
           />
           <label htmlFor="amount-other" className="large-label-size">
-            <span className="currency-symbol-container">{this.props.currencySymbol}</span>
+            <span className="currency-symbol-container">
+              { this.props.currencySymbol ?
+                <span>{this.props.currencySymbol}</span> :
+                <span>&nbsp;</span>
```
alicoding: and here...

components/amount-buttons.jsx
ScottDowne: So if the page is still parsing for the currency/amount/presets, it doesn't display incorrect info or a flat buttons with no height.

Instead, I display nothing, and add in the space so the button has the correct height.

alicoding: Just want to make sure you know you can just do `''` but if you intentionally want a space then this is fine.

components/donate-button.jsx
```diff
@@ -25,9 +25,19 @@ module.exports = React.createClass({
...
+    if (state === "currency") {
+      this.setState({
+        currency: value
+      });
+    }
+  },
   mixins: [IntlMixin],
   render: function() {
-    if (this.state.amount) {
+    if (this.state.amount ) {
```
alicoding: remove this whitespace :P

components/amount-buttons.jsx
ScottDowne: the '' creates a flat button. Need a nbsp to force it to have something.

alicoding:
r+

components/amount-buttons.jsx
ScottDowne: I expected the '' to do that too, then I tried ' ' which I was also annoyed didn't work. Then I tried '&nbsp;' which displayed the text. lol, so adding it to a span not as text fixed it.

PR #836, opened 2015-11-18
Add more strings to messages.properties file

### LINKED ISSUE #835: Localize variants of "this really helps us"
These are the variants:
Get important news and updates
Please allow us to keep in touch
We want you to stay informed
This helps us reach you
This means we can reach you
Don’t miss updates

Need these in:
es, pt-br, it, id, de, fr

alicoding:
closes #835

### DIFF
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -102,6 +102,13 @@ denotes_required_field=denotes required field
 share_a_link_with_your_fellowers=Share a link with your followers
 i_donated_to_mozilla=I donated to @mozilla today because I #lovetheweb. Join me and help build + protect the Web forever
 we_include_you=We want to include you in our work!
+These are the variants:
+get_important_news=Get important news and updates
+please_allow_us_in_touch=Please allow us to keep in touch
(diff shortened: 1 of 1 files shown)


alicoding:
@ScottDowne R?

locales/en-US/messages.properties
```diff
@@ -102,6 +102,13 @@ denotes_required_field=denotes required field
 share_a_link_with_your_fellowers=Share a link with your followers
 i_donated_to_mozilla=I donated to @mozilla today because I #lovetheweb. Join me and help build + protect the Web forever
 we_include_you=We want to include you in our work!
+These are the variants:
```
flodolo: This is broken, it should be a comment. As it is, it's interpreted as an empty en-US string.

alicoding: Ah! Good catch, thanks! I'll fix this asap.

ScottDowne: Sorry, already merged :(

I saw that, and thought it was kinda odd, but I'm not super familiar with .properties syntax and thought maybe it could handle it without the =

PR #839, opened 2015-11-18
Add missing comment tag

yfdyh000:


alicoding:
Thank you so much. Good catch :)

PR #847, opened 2015-11-18
Don't throw on ENOENT

alicoding:


alicoding:
@jbuck R?

jbuck:
@alicoding r+ on travis working

PR #853, opened 2015-11-18
using renderToString instead of renderToStaticMarkup for route pages,…

ScottDowne:
… but not for index


ScottDowne:
@alicoding r?

PR #862, opened 2015-11-18
Add robots tag no not allow index or follow

### LINKED ISSUE #776: Add no-index and no-follow for thank-you page
Add this `<META NAME="ROBOTS" CONTENT="NOINDEX, NOFOLLOW">` to `thank-you` page.

alicoding:
closes #776

We don't want `thank-you` page to be index or follow by search engines, so adding this hopefully those engines will respect that :)

### DIFF
--- a/pages/index.jsx
+++ b/pages/index.jsx
@@ -8,9 +8,13 @@ import Pontoon from '../components/pontoon.jsx';
 var Index = React.createClass({
   render: function() {
     var metaData = this.props.metaData;
+    var robots = 'index, follow';
     var googleFonts = "https://fonts.googleapis.com/css?family=Open+Sans:600,400,300,300italic";
     if (this.props.localeInfo === "cs") {
       googleFonts += "&subset=latin-ext";
+    }
+    if (metaData.current_url.indexOf('thank-you') !== -1) {
+      robots = 'noindex, nofollow';
     }
     var fileHashes = JSON.parse(fs.readFileSync(Path.join(__dirname, '../public/webpack-assets.json')));
     var commentIE = '<!-- [if lt IE 10]><script src="/js/jquery.min.js"></script><script src="/js/placeholder.min.js"></script><![endif]-->';
@@ -20,6 +24,7 @@ var Index = React.createClass({
           <meta charSet="UTF-8"/>
           <meta httpEquiv="X-UA-Compatible" content="IE=edge"/>
           <meta name="viewport" content="width=device-width, initial-scale=1"/>
+          <meta name='robots' content={robots}/>
           <meta property="og:type" content="website" />
           <meta property="og:title" content={metaData.title} />
           <meta property="og:site_name" content={metaData.site_name} />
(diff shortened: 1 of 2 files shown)


alicoding:
@ScottDowne R?

ScottDowne:
Remind me about this tomorrow if I have not yet responded.

scripts/route-file-content.js
```diff
@@ -37,6 +37,7 @@ module.exports = function(outputPath, callback) {
       var contentOfTheFile = React.renderToStaticMarkup(index({
         localeInfo: locale,
         metaData: {
+          current_url: outputPath,
```
ScottDowne: Is this being used anywhere?

pages/index.jsx
```diff
@@ -8,10 +8,14 @@ import Pontoon from '../components/pontoon.jsx';
 var Index = React.createClass({
   render: function() {
     var metaData = this.props.metaData;
+    var robots = 'index, follow';
     var googleFonts = "https://fonts.googleapis.com/css?family=Open+Sans:600,400,300,300italic";
     if (this.props.localeInfo === "cs") {
       googleFonts += "&subset=latin-ext";
     }
+    if (metaData.current_url.indexOf('thank-you') !== -1) {
```
ScottDowne: Can we make this check `>= 1` or `> 0`? The `!== -1`? I think it might be easier to read.

ScottDowne:
Just one nit, consider the fix, then either fix + land, or don't fix + land.

pages/index.jsx
alicoding: #### Checking occurrences

Note that '0' doesn't evaluate to true and '-1' doesn't evaluate to false. Therefore, when checking if a specific string exists within another string the correct way to check would be:

'Blue Whale'.indexOf('Blue') !== -1; // true
'Blue Whale'.indexOf('Bloe') !== -1; // false

Quoted from MDN.

ScottDowne: That's just MDN being lazy. You can also do.
```
'Blue Whale'.indexOf('Blue') >= 0; // true
'Blue Whale'.indexOf('Bloe') >= 0; // false
'Blue Whale'.indexOf('Blue') > -1; // true
'Blue Whale'.indexOf('Bloe') > -1; // false
```
These have the advantages of not being double negatives.

If you were to tell someone this logic in English, would you say "'Blue' is not not in 'Blue Whale'" ? or would you just say "'Blue' is in 'Blue Whale'".

Did MDN give a reason? `> -1` might be the fastest too http://jsperf.com/indexof-test-test-test-test

I think it's ok to question MDN :P

ScottDowne: I'm fine with the fix and merge, now we're just learning stuff.

ScottDowne: You could also flip you variable values, and do a === -1 check. So instead of asking is it not not in the string, you are asking if it's not in the string, which is not not a single negative :P

PR #872, opened 2015-11-19
Fix issue #775 - SEO for H1 tag

### LINKED ISSUE #775: Update page title to "Donate" or "Donate to Mozilla"
Right now our meta og:title is set to `donate.mozilla.org` which is not optimized for SEO. We should update this to `Donate` or `Donate to Mozilla`.

alicoding:
closes #775

### DIFF
--- a/components/header.jsx
+++ b/components/header.jsx
@@ -1,11 +1,22 @@
 import React from 'react';
 
 var Footer = React.createClass({
+  propTypes: {
+    alt: React.PropTypes.string,
+    locale: React.PropTypes.string
+  },
   render: function() {
+    var alt = "Mozilla";
+
+    // FIXME: Should update the list in the regex for locales that did the translation
+    // for whatever `alt` that has been translated.
+    if (/^(en)(\b|$)/.test(this.props.locale)) {
+      alt = this.props.alt;
+    }
     return (
       <div className="header">
         <h1>
-          <img alt={this.props.alt || "Mozilla"} src="/images/mozilla.svg" />
+          <img alt={alt} src="/images/mozilla.svg" />
         </h1>
         <div className="header-copy">
           <div className="row">
--- a/components/header.jsx
+++ b/components/header.jsx
@@ -5,7 +5,7 @@ var Footer = React.createClass({
     return (
       <div className="header">
         <h1>
-          <img alt={this.props.alt} src="/images/mozilla.svg" />
+          <img alt={this.props.alt || "Mozilla"} src="/images/mozilla.svg" />
         </h1>
         <div className="header-copy">
           <div className="row">
--- a/components/header.jsx
+++ b/components/header.jsx
@@ -2,13 +2,9 @@ import React from 'react';
 
 var Footer = React.createClass({
   render: function() {
-    var waterMark;
-    if (!this.props.hideWaterMark) {
-      waterMark = (<img src="/images/mozilla.svg"/>);
-    }
     return (
       <div className="header">
-        {waterMark}
+        <img alt={this.props.alt} src="/images/mozilla.svg" />
         <div className="header-copy">
           <div className="row">
             {this.props.children}
--- a/components/header.jsx
+++ b/components/header.jsx
@@ -4,7 +4,9 @@ var Footer = React.createClass({
   render: function() {
     return (
       <div className="header">
-        <img alt={this.props.alt} src="/images/mozilla.svg" />
+        <h1>
+          <img alt={this.props.alt} src="/images/mozilla.svg" />
+        </h1>
         <div className="header-copy">
           <div className="row">
             {this.props.children}
--- a/components/header.jsx
+++ b/components/header.jsx
@@ -2,9 +2,13 @@ import React from 'react';
 
 var Footer = React.createClass({
   render: function() {
+    var waterMark;
+    if (!this.props.hideWaterMark) {
+      waterMark = (<img src="/images/mozilla.svg"/>);
+    }
     return (
       <div className="header">
-        <img src="/images/mozilla.svg"/>
+        {waterMark}
         <div className="header-copy">
           <div className="row">
             {this.props.children}
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -86,6 +86,7 @@ Bitcoin=Bitcoin
 donation_notice_bitcoin=The Mozilla Foundation is a California non-profit corporation exempt from United States federal income taxation under IRC 501(c)(3) and a public charity classified under IRC sections 170(b)(1)(A) and 509(a)(1). Bitcoin donations Mozilla receives are considered charitable contributions under U.S. federal tax laws, to be used in its discretion for its charitable purposes. If you have questions, check out <a href="https://wiki.mozilla.org/Donate">our FAQ</a> or contact us at <a href="mailto:donate@mozilla.org">donate@mozilla.org</a>.
 email_invalid=Please make sure you have entered a valid email address.
 firefox_footer=Mozilla is the non-profit organization behind Firefox.
+donate_to_mozilla=Donate to Mozilla
 
 # some of these strings are not or will on the page, or some of them will be use for email/snippets only
 this_really_help_us=This really helps us!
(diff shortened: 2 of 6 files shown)


alicoding:
@ScottDowne for code R?

alicoding:
@lovegushwa can you give some feedback on the update with the word `Donate to Mozilla`. I've removed the logo since it looks weird when they're close to each other.

screen shot 2015-11-19 at 3 09 40 pm
screen shot 2015-11-19 at 3 09 51 pm

locales/en-US/messages.properties
```diff
@@ -86,6 +86,7 @@ Bitcoin=Bitcoin
 donation_notice_bitcoin=The Mozilla Foundation is a California non-profit corporation exempt from United States federal income taxation under IRC 501(c)(3) and a public charity classified under IRC sections 170(b)(1)(A) and 509(a)(1). Bitcoin donations Mozilla receives are considered charitable contributions under U.S. federal tax laws, to be used in its discretion for its charitable purposes. If you have questions, check out <a href="https://wiki.mozilla.org/Donate">our FAQ</a> or contact us at <a href="mailto:donate@mozilla.org">donate@mozilla.org</a>.
 email_invalid=Please make sure you have entered a valid email address.
 firefox_footer=Mozilla is the non-profit organization behind Firefox.
+donate_to_mozilla=Donate to Mozilla
```
ScottDowne: What's the process like for this if it's not ready yet in the 5 priority languages?

I only ask because we wanted to ship 25% snippet traffic to our 5 target locales tomorrow.

ScottDowne:
The code looks fine, but I think there are bigger design and l10n questions here.

edrushka:
@alicoding can you clarify what you need the design review on? I can't tell by  your screencaps. Thanks!

ScottDowne:
@edrushka that "donate to mozilla" header bit is added, and the mozilla watermark was removed.

Having that text displayed helps with SEO because google can search our page for the text.

alicoding:
@ScottDowne updated this to only work for `en` locales only for now.

lovegushwa:
@alicoding @ScottDowne does "Donate to Mozilla" persist across all pages of the funnel, or just page 1?

alicoding:
@lovegushwa persist across all pages of the funnel.

edrushka:
Does this need to be in the H1 tag for SEO? We can't add it to the text under the form? Or is  there another way we can accommodate organic searches that isn't a giant H1 string?

lovegushwa:
@alicoding it's not going to work on the thankyou page:

screen shot 2015-11-19 at 1 30 07 pm

lovegushwa:
@ScottDowne @alicoding perhaps this can a separate version of the form. One that does not get most of the traffic, but enough to boost SEO for organic search?

alicoding:
This has to be an H1 and will only be display on donation form page and not
any other pages.

alicoding:
I would suggest we always do it on the form and using h1 and visible on top
of the page. Please don't try to hack it and make it as organic as possible
because we don't want our site to ever be ban by search engine here.

adamlofting:
Excuse my drive by comment, but I suspect there's a lot of trust and value in having the Mozilla logo/wordmark visible on the page when we're asking for money.

alicoding:
We still have the logo on the footer which is visible if you visit the
page?

adamlofting:
But, it's only visible if you're on a fancy desktop/laptop with a high enough resolution display. 

From an SEO point of view, if we set the HTML title tag to start with 'Donate to Mozilla' it will suffice. There are no other pages in Google Results now that would compete with that, on a mozilla domain, with the heading 'donate now' also included on the page.

SEO is going to drive a tiny fraction of total traffic, and the terms we're optimizing for are for people who already want to donate. 

I'm glad we're thinking about SEO, but a small impact on conversion rate for snippet traffic will be much more significant than a small gain from SEO. 

Also, maybe we can A/B test this rather than adding it straight away. My hunch here could be wrong.

alicoding:
I agree with A/B test the location to see the conversion rate but our
search result is still very poor and we are actually competing with a lot
of blogs that have the word donate for Mozilla where most of them is good
but some will drive the user away, but I guess you're only concerning about
the Mozilla domain and the only one I can think of is sendto which
eventually will be redirect to donate.mozilla.org.

Now I've made an argument with Scott and Jon that because our traffic from
search is pretty low and I feel like adding a bit of work to optimize here
is not a bad idea at all as long as we don't affect the conversation rate
of course.

I think what we can do here is still leaving the logo up there but move the
position a bit so it doesn't look weird when they both stay close to each
other.

ScottDowne:
Jbuck raised a good point, we'll be redirecting sendto to donate. What happens then? Does it still stay at the top? Does it go away, or does it slowly move down?

alicoding:
It will stay for a while until a better site content comes up. It's a SEO
game.

valianttry:
Hey @alicoding , let's discuss. This is a significant change and might have consequences. Even subtle changes to our main donation form design have gained or lost $hundreds of thousands USD. I understand the need for a page title to help SEO, but I didn't realize that meant a large H1 font, and a whole new phrase at the top of the form. I thought it was something that could be hidden in the page. 

SEO income in 2014: ~$60,000. Even if we reached $100,000 thanks to SEO --- there could be serious consequences after making a big change like this to the design of the donation form that costs us more than $100,000 in donations. So we could end up losing much more than we gain from SEO improvements.

Q: What do sites do that are heavily image-based? I don't see H1 titles like this on many popular websites. Are you sure this is the only way to optimize for SEO?

alicoding:
Hiding it is a bad idea but what we could do is change the font size to
make it way smaller than what I have. That's why I want design feedback
here because I want to get other people weigh in before we merge this.
Other site like blog platform for instance have this mostly on the very top
of the page to describe their page and that's what H1 really there for.

alicoding:
So after hour of research I have discovered a nice trick here which is going to be a win for everyone in this thread (for me I want SEO real bad ᕙ(⇀‸↼‶)ᕗ) 

What I did here now instead using an `alt` and make an image to be `h1` instead :) So basically our `Mozilla` logo is now an `H1` with an `alt` to the keyword we want :+1:

pages/sequential.jsx
```diff
@@ -78,9 +79,14 @@ module.exports = React.createClass({
...
     }
+    // FIXME: Right now we only display this header for `en` locales
+    // this should be change to display for all once we have other locales translated.
+    if (/^(en)(\b|$)/.test(this.props.locales[0])) {
+      header = (<Header alt={this.getIntlMessage('donate_to_mozilla')}/>);
+    }
     return (
       <div className={className}>
-        <Header/>
+        {header}
```
ScottDowne: Given that the alt isn't displayed to the user, is it safe to say we can remove locale check?

alicoding: Nope because it will still be display on Google search result and I think letting Google pick up the next possible translated strings which could be the word "Donate now" is better than getting the whole `h1` which might not be translated at all.

ScottDowne:
@alicoding I left a comment.

This looks like a good option that doesn't visually change anything that you found, good work.

components/header.jsx
```diff
@@ -4,7 +4,9 @@ var Footer = React.createClass({
   render: function() {
     return (
       <div className="header">
-        <img src="/images/mozilla.svg"/>
+        <h1>
+          <img alt={this.props.alt || "Mozilla"} src="/images/mozilla.svg" />
```
ScottDowne: We should probably add alt to the list of props for this component.

pages/sequential.jsx
```diff
@@ -78,9 +79,14 @@ module.exports = React.createClass({
     if (!this.state.showCvcHint) {
       cvcHintClassName += " hidden";
     }
+    // FIXME: Right now we only display this header for `en` locales
+    // this should be change to display for all once we have other locales translated.
+    if (/^(en)(\b|$)/.test(this.props.locales[0])) {
```
ScottDowne: If we decide to keep this, possible we can build it into the Header component so we don't need to have the same code in two spots? You might have to pass in locale[0] to do the check, but I'm fine with that.

ScottDowne:
Another comment, we can keep the locale check :)

alicoding:
@ScottDowne ready for final review.

PR #876, opened 2015-11-19
Fix issue #875 - only enable de|en for mailing

### LINKED ISSUE #875: Remove signup for other locales than the 2 we have
We currently only have `de` and `en-US` that we can accept for mailing list, but it seems like other locales being enable for this which it shouldn't be.

alicoding:
closes #875


alicoding:
@ScottDowne R?

PR #883, opened 2015-11-20
Optimize imgs and hash them

### LINKED ISSUE #830: Could we reduce the size of the payment logos image?
It's not incredibly large, but I wonder if there are any additional optimizations we can apply.

alicoding:
This patch included image optimization script and hashed all the images
for caching purposes

closes #806 
closes #830


alicoding:
@ScottDowne and/or @jbuck R?

ScottDowne:
@alicoding @jbuck I don't have strong opinions here, so it looks fine to me.

However, I think Jon is going to need to take a look as it's more likely he has opinions on this.

alicoding:
@jbuck I just updated this PR to use hash based on the file like you recommended.

alicoding:
@jbuck anything else on this patch?

Issue #882, opened 2015-11-20
Twitter Social Share is cutoff


valianttry:
This is a really weird image:
![facebook](https://cloud.githubusercontent.com/assets/9830371/11304787/6b8aedf6-8f5e-11e5-9ead-86cccd7de4ce.png)

The sentence is cutoff here:
![share_a_link_on_twitter](https://cloud.githubusercontent.com/assets/9830371/11304766/58ed7402-8f5e-11e5-8f3d-820a4a4dd861.png)

alicoding:
Dup of https://github.com/mozilla/donate.mozilla.org/issues/442

jbuck:
I'm going to turn this issue into fixing the twitter share

valianttry:
Twitter text should be: 

I donated to @mozilla today because I #lovetheweb. Join me and help build + protect the Web forever: https://donate.mozilla.org

jbuck:
@valianttry working on staging: https://donate.mofostaging.net/en-CA/share/

jbuck:
Confirmed fixed on production: https://donate.mozilla.org/en-CA/share/

PR #887, opened 2015-11-20
URI encode Twitter share text

jbuck:


jbuck:
@alicoding r?

jbuck:
Fix #882

alicoding:
R+ once Travis passes

Issue #874, opened 2015-11-19
Donor reports a non-descript error when trying to donate


valianttry:
I searched for him in Stripe, and nothing came up. 

![error_at_your_web_site_for_donation_-_andrea_mozillafoundation_org_-_mozilla_mail](https://cloud.githubusercontent.com/assets/9830371/11283627/7cf85b4a-8eba-11e5-9688-074d56604fdd.png)

jbuck:
I haven't been able to find him either :/

ScottDowne:
I cannot find it either, I found one that was close, but it ended up being this morning, which means after this email.

That error comes when we cannot make a connection to stripe, so maybe there was something blocking stripe from the network? Just a guess.

jbuck:
Scott and I looked at the possible Stripe errors and we found that there are some communication errors we're not handling yet. Scott is going to add error messages for those.

ScottDowne:
It's most likely a "api_connection_error 	Failure to connect to Stripe's API."

I'm thinking we can display a message for that, but I don't want to use that message...

We can also do something like: "Something seems to have gone wrong. Please try again later. [api_connection_error]"

This way if stripe sends us an unknown error and the user screen shots it, we can have a bit more info to debug.

jbuck:
@ScottDowne Maybe ask the user if they have any blocking scripts running or something?

ScottDowne:
@jbuck or under a public network?

ScottDowne:
I'm thinking we need a specific message for connection error.

Something like "Sorry there was a connection error. Please try again later."

Thoughts?

valianttry:
"Something seems to have gone wrong. Please try again later. [api_connection_error]"
Would be my preference, since "connection error" could signal they need to plug their ethernet cord in or something.

edrushka:
Just noting that when I had my "rotating cog forever" problem, it was also using Mastercard.

PR #896, opened 2015-11-20
adding error code to error message if it's a generic error.

ScottDowne:


ScottDowne:
I could test this by making it produce an [abort] error code at the end of the error message.

I couldn't get any other error message to be produced, but, this is at least something to help us debug.

@alicoding r?

ScottDowne:
Fixes #https://github.com/mozilla/donate.mozilla.org/issues/874

ScottDowne:
@jbuck r?

mixins/form.jsx
```diff
@@ -193,7 +193,7 @@ module.exports = {
     if (errorType === "card_error" && cardError) {
       form.error(cardError.field, cardError.message);
     } else {
-      form.error("other", this.getIntlMessage('try_again_later'));
+      form.error("other", this.getIntlMessage('try_again_later') + " [" + errorType + "]");
```
alicoding: Does errorTpye needs to be localized?

ScottDowne: Nah, it's mostly for us, and it's basically an error code.

alicoding: You're saying that this is not going to be visible to the user or on the page?

jbuck: It will be user visible, but the idea is that if someone takes a screenshot we can easily understand the error

alicoding: Well hoping the user will do that? Some might but some would just leave if they don't understand if this is visible to user in different language?? I just want to see if we can localize this

jbuck: I don't want to localize this, because I want to be able to see a screenshare and know exactly what kind of error occured even if the language is different. This is for us, not our users

ScottDowne: @alicoding it's not going to be an error message, it's an error type. So it's [api_connection_error]

What would that translate to?

ScottDowne: I also understand we won't get many of these screenshots, but, we only need a couple to fix it for the rest.

alicoding: So, we are going to get this

`Something seems to have gone wrong. Please try again later ["api_connection_error"]`.

This is fine, but if we are hoping if any user (donors) to report back an error then surely we need a better message here. (I won't send a screenshot if it happened to me when trying to donate somewhere LOL).

ScottDowne: This is in response to an actual screenshot in an email we got, which we couldn't fix because we knew nothing about it.

I wouldn't have sent it either.

ScottDowne:
@alicoding r?

Issue #912, opened 2015-11-22
Reduce default amount on snippet >> donation form for BRL


valianttry:
The PTBR form has a relatively low conversion rate of 7 - 8% compared to other locales. I suggest we reduce the default selected amount on the snippet and donation form to R$25 (about $6 USD). Right now it's R$50 - or about $14 USD. 

![about_home_snippet_preview](https://cloud.githubusercontent.com/assets/9830371/11321448/96dd8b3c-9074-11e5-841a-4c54edc0b53e.png)

![donate_mozilla_org___doe_a_mozilla_hoje](https://cloud.githubusercontent.com/assets/9830371/11321449/9d0a3e92-9074-11e5-819e-e8f60107b3c2.png)

valianttry:
closing this, Jean made the change so it is now: 50, 25*, 15, 10

ScottDowne:
Actually, we just changed which was selected, we didn't update the presets themseelves. So it's now 100, 50, 25*, 10.

https://snippets.mozilla.com/show/5487/

adamlofting:
See also #518

PR #913, opened 2015-11-22
Lower BRL defaults

jbuck:


ScottDowne:
@jbuck Does this have an issue attached to it with more info?

Single the same as monthly? It doesn't really matter, as snippet is going to basically handle this anyway, but, we have them as a bit different in every other currency. Sounds like you feel we shouldn't be bothering with different presets for frequency.

jbuck:
Just #912. The snippet data isn't permanent, I want to make sure we update our coded defaults so our platform is in the best possible shape to accept donations throughout the year. Not sure about whether we should have different monthly/one-time defaults

ScottDowne:
Would you be against setting single to 100, 50, 25, 10 and monthly to what you have until we have better data?

PR #919, opened 2015-11-23
Update Facebook share image

alicoding:


alicoding:
@ScottDowne R?

ScottDowne:
@alicoding This looks good to me.

PR #920, opened 2015-11-23
Improve logging - good-console-logfmt + stripe&paypal logging + ops logging

### LINKED ISSUE #769: How can improve our logging?
Right now we have a simple logger that logs every response in https://github.com/mozilla/donate.mozilla.org/blob/6ca4669ae9d4e0ded144dc224401442ec8be10f1/server.js#L211-L222

This is useful for local development, but not very useful for running on Heroku. Heroku already shows the response logs, so I don't think we need to display that. I think it'd be useful to print errors to the logs. We should also consider printing ops information to the logs.

cadecairos:
Fixes #769 

@jbuck R?

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "extract-text-webpack-plugin": "0.9.1",
     "fs-writefile-promise": "1.0.3",
     "good": "6.4.0",
-    "good-console": "5.2.0",
+    "good-console-logfmt": "^1.0.0",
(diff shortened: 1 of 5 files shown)


cadecairos:
@jbuck You might also want to check out the changes I made to good-console when I forked it: https://github.com/hapijs/good-console/compare/master...cadecairos:master

ScottDowne:
Looks good to me. R+

PR #923, opened 2015-11-23
allow for input of comma in amount button, and display this value in …

ScottDowne:
…the proper locale if it came from query string

### DIFF
--- a/components/amount-buttons.jsx
+++ b/components/amount-buttons.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import IntlMixin from 'react-intl';
 import {FormattedMessage, FormattedNumber} from 'react-intl';
 import {ErrorMessage} from './error.jsx';
 import listener from '../scripts/listener.js';
@@ -34,6 +35,7 @@ var AmountButton = React.createClass({
 });
 
 var AmountOtherButton = React.createClass({
+  mixins: [IntlMixin],
   propTypes: {
     checked: React.PropTypes.bool.isRequired,
     onRadioChange: React.PropTypes.func,
@@ -41,12 +43,12 @@ var AmountOtherButton = React.createClass({
     amount: React.PropTypes.string,
     currencySymbol: React.PropTypes.string,
     placeholder: React.PropTypes.string,
-    locale: React.PropTypes.string.isRequired,
     userInputting: React.PropTypes.bool
   },
   getInitialState: function() {
     return {
-      inputValue: ""
+      inputValue: "",
+      decimalCurrency: this.testNumberFormat()
     };
   },
   onRadioClick: function() {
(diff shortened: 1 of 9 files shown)


ScottDowne:
WIP

ScottDowne:
Needs tests.

ScottDowne:
Added tests and I think karma should pass. Waiting for travis to go green.

@alicoding can I get a really thorough review?

components/amount-buttons.jsx
```diff
@@ -53,7 +60,36 @@ var AmountOtherButton = React.createClass({
       this.props.onRadioChange();
     }
   },
+  onInputChange: function(e) {
+    var inputValue = e.currentTarget.value;
+    var amount = "";
+
+    if (/^(\d)*[\,]?(\d){0,2}$/.test(inputValue)) {
```
alicoding: Can you add a comment above this regex that you are trying to capture as many number follow by comma and follow by 0 to 2 numbers?

components/amount-buttons.jsx
```diff
@@ -53,7 +60,36 @@ var AmountOtherButton = React.createClass({
...
+      amount = inputValue.replace(",", ".");
+    } else if (/^(\d)*[\.]?(\d){0,2}$/.test(inputValue)) {
```
alicoding: And here.

components/amount-buttons.jsx
alicoding: Actually I think this test is not enough. We need to consider about locales that uses comma as a separator like in the US and Canada e.g. 2,000.00 How are we going to deal with this?

alicoding: My suggestion would be to check the currency/locale here as well.

components/amount-buttons.jsx
alicoding: Same goes for this case. 2.000,00 ?

components/amount-buttons.jsx
```diff
@@ -53,7 +60,36 @@ var AmountOtherButton = React.createClass({
...
+  },
   render: function() {
+    var amount = this.props.amount;
+    var inputValue = this.state.inputValue;
+    if (!amount) {
+      inputValue = "";
+    } else if (!this.props.userInputting) {
+      // We only need this for initial display before the user starts inputting,
+      // once they start inputting, we can adapt.
+      inputValue = new Intl.NumberFormat(this.props.locale, {
```
alicoding: Can't you use `this.formatNumber()` instead here?

alicoding:
We need more test case for the case of 2,000.00 vs 2.000,00

components/amount-buttons.jsx
ScottDowne: Yes

components/amount-buttons.jsx
ScottDowne: It makes turning 1,00 -> 1.00 much harder. I need to start parsing and I kinda wanted to attempt a solution first that didn't parse.

ScottDowne: Right now adding a dot or comma can only happen once, and only before the last two characters, making it really easy to parse.

alicoding: does it switch based on the locale?

ScottDowne:
@alicoding Updated, travis is running, should be green. R?

Issue #829, opened 2015-11-18
Could we reduce the number of fonts we use?


jbuck:
Right now we're loading 4 fonts: 1 Font Awesome & 3 Open Sans. Is it possible to reduce this?

PR #931, opened 2015-11-25
25s timeouts for paypal and stripe requests

cadecairos:
#829 

@jbuck R?

### DIFF
--- a/routes/paypal.js
+++ b/routes/paypal.js
@@ -1,6 +1,10 @@
-var httpRequest = require('request');
+var request = require('request');
 var querystring = require('querystring');
 var paypalLocales = require('../intl-config.js').paypalLocales;
+
+var httpRequest = request.defaults({
+  timeout: 25000
+});
 
 function setupPaypal(transaction, recurring, callback) {
   var charge = {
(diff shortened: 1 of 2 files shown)


ScottDowne:
@cadecairos or @jbuck when it times out, the user gets a generic error?

cadecairos:
@ScottDowne It should get a https://httpstatuses.com/408 instead of a https://httpstatuses.com/503

alicoding:
Can I test this somehow?

ScottDowne:
@alicoding maybe with no network.

cadecairos:
@alicoding you could use an intercepting proxy like ZAProxy, intercept requests to paypal/stripe and then let the request timeout.

shameless plug: https://chrisdecairos.ca/intercepting-traffic-with-zaproxy/

jbuck:
@cadecairos this looks good, but I'm curious what happens the users hit this error... do they just get a "try again" message?

cadecairos:
No, they get json, because this is a redirect. that's why I keep saying we need a landing page for errors!

jbuck:
As shitty as that is, people currently get a nice 503 page from Heroku....

PR #932, opened 2015-11-25
add debit to credit card button. Now that the new length is longer, r…

ScottDowne:
…emove margins to make space for new length of copy and properly center the images above the copy which now makes it look not centered.


ScottDowne:
@alicoding r?

PR #933, opened 2015-11-25
add a bit of space to buttons for mobile view

ScottDowne:


ScottDowne:
@alicoding r?

In mobile view, the amount buttons with huf currency don't fit in the buttons. I fixed this by reducing the font, which makes sense for mobile, but not desktop.

PR #946, opened 2015-11-26
Disable and hide postal code input if the country doesn't have them

### LINKED ISSUE #925: erm... apparently there are no postal codes in Bolivia?
More of an "FYI" as we have received $0 from Bolivia (but maybe this is part of the reason why?)

i_tried_to_donate__10u_s__-_andrea_mozillafoundation_org_-_mozilla_mail

cadecairos:
Fixes #925

### DIFF
--- a/components/address-input.jsx
+++ b/components/address-input.jsx
@@ -747,7 +747,7 @@ var Code = React.createClass({
       return;
     }
 
-    if (COUNTRIES_WITHOUT_POSTCODES.includes(detail.value)) {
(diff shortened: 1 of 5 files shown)


cadecairos:
@ScottDowne / @alicoding I can't figure out how to test this - There doesn't appear to be a way to change react component state in the test utils?

cadecairos:
Oh. "updateState" how nice

cadecairos:
Nope, doesn't work. the component isn't mounted, and can't have set state called

jbuck:
We're switching to the one-page form so we don't need to land this

PR #947, opened 2015-11-26
add route for /assets/{params*}

alicoding:
We need to make sure that this route can still look for files that CDN could give us some :hankey:


alicoding:
@jbuck R?

jbuck:
@alicoding r+

PR #963, opened 2015-11-27
switch default page from sequential to one-page

ScottDowne:


ScottDowne:
@alicoding or maybe @jbuck r?

Issue #936, opened 2015-11-25
other amount currency symbol is no longer centered


ScottDowne:
![other-symbol](https://cloud.githubusercontent.com/assets/197334/11407960/fd07e8dc-9383-11e5-8133-ab71230a1e30.png)

Has this always been like this, or did we break something?

KasperChristensen:
https://github.com/mozilla/donate.mozilla.org/commit/937652fd5b3d1df61109dd265f1d1ad360348d3f#diff-c5786860e11375aed83e0b24e313690e

It changed here:
in less/shared.less padding: 6px 12px; is changed to padding: 6px 0;

ScottDowne:
Interesting, that's exactly when I noticed it break too.

ScottDowne:
Boom fixed! closing! Thanks @KasperChristensen

Just in time for a production ship! Should be live in a few minutes.

valianttry:
@KasperChristensen it's so awesome to see a contributor's help with fundraising! Thank you so much! :+1:

PR #965, opened 2015-11-27
#936 changed width of currency-symbol-container so it centers again

KasperChristensen:


ScottDowne:
This input has two styles, which we plan on testing very soon: https://donate.mozilla.org/en-US/?test=new-flow-test

I noticed it's centered, and it centers it with this line here: https://github.com/mozilla/donate.mozilla.org/blob/master/less/new-flow-test/amount-buttons.less#L4

However, I think I like your approach more, so we should remove the `12px` in `padding: 5px 12px;` and just go with `padding: 5px 0;` and let the 60px width shine through. Thoughts?

ScottDowne:
Actually, we can merge this and just fix the test flow in another patch!

PR #968, opened 2015-11-27
Locale negotiation

### LINKED ISSUE #927: Locale negotiation
Our current UX/design for displaying the locale is like this:

1. If the user visit the URL without locale in it e.g. [https://donate.mozilla.org/share/](https://donate.mozilla.org/share/)
  We will check their locale in `navigator.language` and if we support that language we will redirect them to that locale and replace the URL with the new one e.g. `navigator.language='de'` then the URL will now be [https://donate.mozilla.org/de/share/](https://donate.mozilla.org/de/share/).

2. If the user visit the URL with the locale in it [https://donate.mozilla.org/de/share/
(https://donate.mozilla.org/de/share/).
  We will do nothing here since they intentionally asking for that locale.

Now this design is okay, but what if the case for number 1 we don't support that locale? Right now it will just redirect to our default locale which is `en-US`. This is fine, but what if we do support the locale they've requested, but in another version? The example for this case would be:

List of currently supported locale in our app (this is just some of them)

```
de
it
fr
en-US
en-GB
pt-BR
pt-PT
```

So, if the user visit the site with `navigator.language=pt` what's going to happen here? Right now it will be redirect to `en-US` instead of choosing one of the `pt-*` for them. Another thing we also need to implement here is when there are multiple languages set in the user's language preferences. 

1. If the user set their languages to `navigator.languages=['de','fr','en-GB']`.
  What we need to do here is check the first item in the array of `navigator.languages` and redirect to the first possible locale, and if we don't support that locale then we'll move to the next item until we find a possible match supported locale otherwise will default to `en-US`. The example above is simple because we do support all 3 languages.

2. If the user set their languages to `navigator.languages=['he','fr-FR','en']`.
  Now the first language we currently don't support, but `fr-FR` is a possible locale to match with our current support language `fr` so we could redirect them to that locale.

note: 

`navigator.language` is whatever you have set your first language in your browser preferences. (Chrome doesn't work well with this option.)
`navigator.languages` is the list of languages in order in your browser preferences.

alicoding:
closes #927


alicoding:
@jbuck R?

alicoding:
@jbuck so I've decided to remove that dependency-check module because it doesn't seem to support ES6 https://github.com/maxogden/dependency-check/issues/37 and it seems to be blocked on another dependency here. I think we don't rely on this as much now since we have bunch of other ways to test to make sure we include module in package.json such as travis.

I also remove that extra build step for those path with no locales now since we do catch them server side, so we don't really need an empty file on /public anymore.

server.js
```diff
@@ -269,7 +269,9 @@ module.exports = function() {
     if (process.env.NPM_CONFIG_PRODUCTION === 'true') {
       server.route(require('./lib/hashed-file-routes')(securityConfig));
     }
-
+    server.route(require('./lib/no-locale-pages.js')(securityConfig, []));
+    server.route(require('./lib/with-locales-pages.js')(securityConfig, []));
+    server.route(require('./lib/unsupported_locales.js')(securityConfig, []));
     server.route([{
       method: 'GET',
       path: '/{params*}',
```
jbuck: I *think* we can remove this

lib/no-locale-pages.js
```diff
@@ -0,0 +1,28 @@
...
+        var query = '';
+        var header = Parser.parse(request.headers["accept-language"]);
+        var languages_array = header.map(function(l) {
+          return l.code + (l.region ? "-" + l.region : "");
+        });
+        if (request.url.search) {
+          query = request.url.search;
+        }
+        reply.redirect('/' + bestLang(languages_array, supported_locales, 'en-US') + page + query)
+        .vary("Accept-Language").header("Last-Modified", lastModified);
```
jbuck: I'd remove the Last-Modified header here

lib/no-locale-pages.js
```diff
@@ -0,0 +1,28 @@
...
+      }
```
jbuck: Need to add 7 day cache-control header here

lib/unsupported_locales.js
```diff
@@ -0,0 +1,28 @@
...
+        var header = Parser.parse(request.headers["accept-language"]);
+        var languages_array = header.map(function(l) {
+          return l.code + (l.region ? "-" + l.region : "");
+        });
+        if (request.url.search) {
+          query = request.url.search;
+        }
+        reply.redirect('/' + bestLang(languages_array, supported_locales, 'en-US') + page + query)
+        .vary("Accept-Language").header("Last-Modified", lastModified);
+      }
```
jbuck: Add a 1 day cache-control header here, so if we add a new language it doesn't take long to be enabled in the redirect

server.js
```diff
@@ -264,29 +266,10 @@ module.exports = function() {
       throw err;
     }
 
-    // We have these routes specifically for production where it's possible that
-    // a CDN index.html may refer to an outdated CSS/JS file that doesn't exist
-    if (process.env.NPM_CONFIG_PRODUCTION === 'true') {
```
alicoding: By removing the `/{params*}` now it requires that we enable this for both `development` and `production` for serving static files.

alicoding:
@jbuck I've added tests for this patch \o/

jbuck:
This requires #1131 to land first to get the tests passing.

alicoding:
@jbuck all tests are passing now \o/

package.json
```diff
@@ -17,11 +17,11 @@
     "test:karma": "karma start",
     "test:coverage": "KARMA_ENV=coverage karma start --reporters coverage,coveralls",
     "watch:test": "karma start --no-single-run --browsers Firefox",
-    "test:http": "mocha tests/http/*.test.js",
+    "test:http": "NPM_CONFIG_PRODUCTION=true mocha tests/http/*.test.js",
```
alicoding: @jbuck I don't know if you can think of another approach here, but this seems to be the cleanest? :P

components/client.jsx
```diff
@@ -1,18 +1,10 @@
 import React from 'react';
 import Router from 'react-router';
 import routes from './routes.jsx';
-import langURLParser from '../scripts/langURLParser.js';
 import queryParser from '../scripts/queryParser.js';
 
 Router.run(routes, Router.HistoryLocation, function(Handler, state) {
   var queryString = state.query;
-  var pathname = langURLParser(state);
```
alicoding: We no longer need client side redirect \o/

alicoding:
@jbuck I think this is now solid and ready for final round of review :)

tests/lib/langURLParser.tests.js
```diff
@@ -32,16 +32,4 @@ describe('langURLParser.js', function() {
   it('should return "/en-US/" if given path "/"', function() {
     should(langURLParser({pathname: '/'})).equal('/en-US/');
   });
-  it('should return "/en-US/thank-you/" if given path "/?redirect=/thank-you/"', function() {
```
jbuck: This entire file should go away, no?

alicoding: \o/

components/routes.jsx
```diff
@@ -3,36 +3,19 @@ import React from 'react';
...
-    {
-      Object.keys(pages).map(function(key) {
+      Object.keys(pages).map(function(page, key) {
         var routeOBJ = {
-          key:  pages[key].name,
-          name: pages[key].name,
-          path: '/' + pages[key].name + '/?',
-          handler: pages[key].handler
+          key:  key,
+          path: pages[page].pathname,
```
jbuck: Forgive me for not being familiar with `react-router` here, but why are you able to remove the `name` key?

scripts/paths.js
```diff
@@ -1,28 +1,11 @@
-import locales from '../locales/index.js';
-import pages from '../data/pages.js';
-var localeString = ':locale';
+var locales = require('../locales/index.js');
```
jbuck: I feel like this file should live in the `data/` directory, since it's not actually an executable script

components/routes.jsx
alicoding: https://github.com/rackt/react-router/blob/0.13.x/doc/01%20Route%20Configuration/Route.md#name-optional

scripts/paths.js
alicoding: Do you want me to do it in this patch, or?

jbuck: I'd do it in this patch. And to be clear, I'm talking about the 'scripts/paths.js' file, not the locale data

components/small-print.jsx
```diff
@@ -3,9 +3,12 @@ import Link from './link.jsx';
 import { FormattedHTMLMessage, FormattedMessage, IntlMixin } from 'react-intl';
 
 var Footer = React.createClass({
+  propTypes: {
+    locale: React.PropTypes.string
```
jbuck: This property is missing from `pages/partial-page.jsx`

lib/build-urls.js
```diff
@@ -0,0 +1,74 @@
...
+  Object.keys(pages).forEach(function(path) {
+    returnPages.push({
+      method: 'GET',
+      path: path,
+      handler: function(request, reply) {
+        processPage(request, reply, path);
+      },
+      config: {
+        cache: {
+          expiresIn: 300 * 1000, // one week
```
jbuck: This is 5 minutes (300 seconds) not one week

lib/build-urls.js
```diff
@@ -0,0 +1,74 @@
...
+  });
+  var query = '';
+  if (request.url.search) {
+    query = request.url.search;
+  }
+  reply.redirect('/' + bestLang(languages_array, locales, 'en-US') + page + query)
+  .vary("Accept-Language");
+}
+
+module.exports = function(securityConfig) {
```
jbuck: This parameter is unused

lib/build-urls.js
```diff
@@ -0,0 +1,74 @@
...
+function processPage(request, reply, page) {
+  var header = Parser.parse(request.headers["accept-language"]);
+  var languages_array = header.map(function(l) {
+    return l.code + (l.region ? "-" + l.region : "");
+  });
+  var query = '';
+  if (request.url.search) {
+    query = request.url.search;
+  }
+  reply.redirect('/' + bestLang(languages_array, locales, 'en-US') + page + query)
```
jbuck: Is it safe to append these strings together like this?

lib/build-urls.js
jbuck: Though it brings up the question of how long a redirect should last for. Longer periods of time are good for cache hit, but what if we added a new supported language?

package.json
jbuck: Hrm, it's not great that you need to set this. I filed https://github.com/mozilla/donate.mozilla.org/issues/1138

lib/build-urls.js
alicoding: I added `url.format()` around that. Not sure if that's enough?

server.js
```diff
@@ -239,11 +241,31 @@ module.exports = function(options) {
...
-      return reply.redirect('/?redirect=' + request.url.pathname);
+      if (pathname && paths.findIndex(function(path) { return path.pathname === '/' + pathname + '/'; }) !== -1) {
+        return reply.redirect('/' + pathname + '/' + query);
+      }
+      return reply.redirect('/' + query);
     }
     return reply.continue();
   });
+  server.ext('onRequest', function(request, reply) {
+    var query = '';
```
jbuck: I feel like there's an easier way to implement this logic, I'm just not sure how...

server.js
```diff
@@ -239,11 +241,31 @@ module.exports = function(options) {
   // This will catch all 404s and redirect them to root URL
   // with preserving the pathname for client-side to handle.
   server.ext('onPreResponse', function(request, reply) {
+    var url = request.url.pathname;
```
jbuck: Gragh this bothers the heck out of me. I wish we had a proper 404 page to serve up.

tests/http/locales.test.js
```diff
@@ -0,0 +1,115 @@
...
+        should(response.statusCode).equal(200);
+        should(response.headers['content-type']).equal(type);
+        should(response.headers['cache-control']).equal(cache);
+        done();
+      });
+    });
+  });
+});
+
+describe('paths with locale', () => {
```
jbuck: What's the difference between this set of tests, and the above set?

tests/http/locales.test.js
```diff
@@ -0,0 +1,115 @@
...
+      cache,
+      equal: '/en-US/',
+      headers: {
+        'Accept-Language':"en;q=0.5"
+      }
+    },
+    '/wx/about/': {
+      type,
+      cache,
+      equal: '/fr/about/',
```
jbuck: Why wouldn't this redirect to en?

tests/http/locales.test.js
```diff
@@ -0,0 +1,115 @@
...
+  };
+
+  Object.keys(noLocalePaths).forEach((key) => {
+    it(`should redirect to ${noLocalePaths[key].equal} when accessing ${key}`, (done) => {
+      instance.inject({
+        url: key,
+        headers: noLocalePaths[key].headers
+      }, (response) => {
+        should(response.statusCode).equal(302);
+        should(response.headers.location).equal(noLocalePaths[key].equal);
```
jbuck: Add a check for the `Vary: Accept-Language` header too

tests/http/locales.test.js
alicoding: I think I can remove these because above sets pretty much covers everything, but I just thought I want hardcoded ones in case the array above didn't work somehow?

server.js
jbuck: Yeah, I think this needs to be improved somehow...

server.js
```diff
@@ -239,11 +242,31 @@ module.exports = function(options) {
...
+    var pathname = url.split('/')[2];
```
jbuck: What happens if I go to something like `/asdsadsadsadasd`

jbuck:
I just have some concerns about the server extension bits of code

server.js
alicoding: In this patch or file for follow up?

jbuck: Lets use http://hapijs.com/api#servermatchmethod-path-host here

server.js
```diff
@@ -13,10 +13,13 @@ var Path = require('path');
 var Hapi = require('hapi');
 var Hoek = require('hoek');
 var Joi = require('joi');
+var Url = require('url');
```
jbuck: not being used

server.js
```diff
@@ -13,10 +13,13 @@ var Path = require('path');
...
 var polyfillio = require('polyfill-service');
 var PolyfillSet = require('./scripts/PolyfillSet.js');
 var exchangeRates = require('./assets/exchange-rates/latest.json');
 var routes = require('./routes');
+var paths = require('./data/paths.js');
```
jbuck: Not being used

server.js
```diff
@@ -13,10 +13,13 @@ var Path = require('path');
...
+var pages = require('./data/pages.js');
```
jbuck: Not being used

jbuck:
@alicoding r+ with the nits I noted in `server.js`

PR #978, opened 2015-11-28
Make polyfill.js more cachable

jbuck:


alicoding:
I r+ for the code, but will run it once I get home if you want me to check as well.

PR #979, opened 2015-11-28
Use official Russian ruble currency symbol instead of abbreviation

unghost:
Use official Russian ruble currency symbol instead of abbreviation, see https://en.wikipedia.org/wiki/Russian_ruble#Currency_symbol


alicoding:
That's great @unghost ! thank you so much for your PR :+1:

PR #994, opened 2015-11-30
Fix issue #991 - add ga for /share page

### LINKED ISSUE #991: Is the redirect to 'share' done via React?
@ScottDowne @alicoding 

We're not seeing much GA traffic reported for our social share page, but I think this is to do with React affecting GA tracking. 

For context: 

This is thank you page views
screenshot 2015-11-30 14 46 29

This is share views (according to GA)
screenshot 2015-11-30 14 46 55

I suspect React is updating the DOM, and the URL, but not triggering a full reload of the page. Would that make sense? 

I was setting up an Optimizely test here, but think we might need to add some custom code to trigger the GA tracking and to start the Optimizely experiment.

alicoding:
closes #991

### DIFF
--- a/pages/share.jsx
+++ b/pages/share.jsx
@@ -4,9 +4,17 @@ import Header from '../components/header.jsx';
 import Social from '../components/social.jsx';
 import ThankYouHeader from '../components/thank-you-header.jsx';
 import { IntlMixin } from 'react-intl';
+import reactGA from 'react-ga';
 
 var ThankYou = React.createClass({
   mixins: [IntlMixin],
+  componentDidMount: function() {
+    // We are using `transitionTo()` in /thank-you page, so when the user
+    // finished the signup and being redirect here it doesn't get track because
+    // react-router doesn't do a full page reload from /thank-you to /share page.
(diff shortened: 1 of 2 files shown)


alicoding:
@adamlofting R?

pages/share.jsx
```diff
@@ -4,9 +4,17 @@ import Header from '../components/header.jsx';
...
+import reactGA from 'react-ga';
 
 var ThankYou = React.createClass({
   mixins: [IntlMixin],
+  componentDidMount: function() {
+    // We are using `transitionTo()` in /thank-you page, so when the user
+    // finished the signup and being redirect here it doesn't get track because
+    // react-router doesn't do a full page reload from /thank-you to /share page.
+    var currentPage = window.location.pathname;
+    reactGA.pageview(currentPage);
```
ScottDowne: @adamlofting would this track it twice for cases where the page is loaded directly?

We think we could put the tracking code just before the transitionTo if it's an issue.

adamlofting:
@alicoding R+

PR #996, opened 2015-11-30
remove stripe.js, ask browser to load stripe-checkout async

### LINKED ISSUE #818: Should we be loading stripe.js and stripe checkout?
My understanding is that we're using stripe.js for the sequential donation form and stripe checkout for the one page donation form. Would it be possible to only load one or the other based on what page is being rendered?

cadecairos:
Fixes #818 

@ScottDowne R?

### DIFF
--- a/pages/single-form.jsx
+++ b/pages/single-form.jsx
@@ -91,6 +91,7 @@ var SingleForm = React.createClass({
         </div>
         <SmallPrint stripeNotice={true} />
         <Footer/>
+        <script async src="https://checkout.stripe.com/checkout.js"></script>
       </div>
     );
   }
--- a/pages/single-form.jsx
+++ b/pages/single-form.jsx
@@ -91,7 +91,7 @@ var SingleForm = React.createClass({
         </div>
         <SmallPrint stripeNotice={true} />
         <Footer/>
-        <script async src="https://checkout.stripe.com/checkout.js"></script>
+        <script src="https://checkout.stripe.com/checkout.js"></script>
       </div>
     );
   }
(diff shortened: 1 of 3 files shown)


pages/single-form.jsx
```diff
@@ -91,6 +91,7 @@ var SingleForm = React.createClass({
         </div>
         <SmallPrint stripeNotice={true} />
         <Footer/>
+        <script async src="https://checkout.stripe.com/checkout.js"></script>
```
ScottDowne: If this is async, are we sure stripe is going to be ready when we go to use it?

jbuck: The user would need to select an amount, click the privacy checkbox, and then the donate checkbox. If you're worried about that case we could add an existence guard to the StripeCheckout.configure() block

ScottDowne: Well, the amount is likely prefilled via snippet.

So it's privacy then donate.

jbuck: So you don't want to make this script async? r-?

ScottDowne: Can we be sure that stripe is ready when we use it?

ScottDowne: I was going to try https://github.com/terinjokes/promisescript with this.

jbuck: No, as the code is currently implemented there is no check

jbuck: So, it'd be nice to leave this inline so the browser's preloader can start loading this as soon as possible, but it's not required to complete the first render of the page, hence async

ScottDowne: How do you envision a check? Would it bail or queue or something?

PR #1004, opened 2015-12-01
add a donation page with more info

ScottDowne:


ScottDowne:
@alicoding r?

data/pages.js
```diff
@@ -4,6 +4,11 @@ var pages = {
     path: '/:locale/',
     handler: require('../pages/single-form.jsx')
   },
+  '/additional-info/': {
```
alicoding: I think we should pick a better URL here? or is this decided?

alicoding:
You will need to do the hashes thing for the image here.

Install

https://github.com/sindresorhus/hasha-cli

Usage:

```
hasha <image> --algorithm=md5
```

Once you've done that you will get a hash and just rename the file including that hash from hasha.

Then go to: /lib/hashed-file-routes.js and add that image in the `files` object.

PR #1021, opened 2015-12-02
Revert "remove stripe.js, ask browser to load stripe-checkout async"

jbuck:
Reverts mozilla/donate.mozilla.org#996


jbuck:
Scott and I were both seeing "StripeCheckout is not defined" errors, so backing this out until we can understand the problem better

PR #1032, opened 2015-12-03
create test for this really helps us arrow and image

ScottDowne:


ScottDowne:
Test url is http://localhost:3000/en-US/?test=this-really-helps

@alicoding r?

alicoding:
You didn't hash bunch of other images. Is that intentional?

alicoding:
This looks good, but only for en-US right now?

ScottDowne:
Yeah, only for en-US, the rest had UI issues but we only needed en-US to test.

I wanted a smaller patch.

ScottDowne:
@alicoding and yes, I didn't bother to hash the images we were not using.

alicoding:
@ScottDowne I would say just hash them, so when we have to use them in the future we can just grab and go :)

ScottDowne:
@alicoding nah, file it. I'll do it later.

alicoding:
r+ then

ScottDowne:
Lol, thanks :)

Issue #998, opened 2015-11-30
Send receipt emails to recurring donors that didn't receive one


jbuck:
In https://github.com/mozilla/donate.mozilla.org/issues/972 I fixed the receipt emails so they work again, but @valianttry wants to go through the donor experience to make sure it's good with the delayed email. We should figure out:

- How many people were affected?

valianttry:
Other Qs:
-- What locales? (in case we have to localize any new content)
-- Did they receive any correspondence at all from us? (e.g. did they receive a PayPal receipt or Stripe receipt?)
-- What will we send them? Just a receipt and it'll appear to be delayed? Will we send all of the owed receipts at the same time? What's the deadline -- e.g. when will they get their 2nd receipt for their 2nd monthly gift? We should send their first one well ahead of the second one.

jbuck:
* All locales are already translated, thankfully enough
* They wouldn't have received any correspondence from us unless they signed up for the mailing list
* Yes, we'd just attempt to resend the email receipt. There's only one email backlog, for donations between when we started the stripe account and yesterday.
* The deadline would be... December 9th? Hrm, shit.

valianttry:
According to Google Analytics we're talking about 1,622 individuals. 
Plan A:
How long before we can send these donors their first receipt? If it's something we can do really soon, like before Mozlando, let's do that!
Plan B:
Q: IS it possible to combine receipts for donation #1 with the donation #2 receipt email? E.g. it would look something like:
---------
TRANSACTION RECEIPT FOR YOUR RECORDS:
Roger Rabbit
Monthly gift amount: $3 (11/09/2015)
Monthly gift amount: $3 (12/09/2015)
Transaction ID: ch_12345abcdefjghipop
1234 American Dream Blvd Apt. F
Big City 33134
United States

valianttry:
whoa. crazy formatting glitch.

jbuck:
I like plan A better!

jbuck:
Charges before 2015-11-28T03:50:58Z

jbuck:
Done!

PR #1035, opened 2015-12-03
A script for sending reciepts for recurring monthly donations on stri…

cadecairos:
…pe (that occurred before a specified data)

#998 

`BEFORE_DATE="2015-11-28T03:50:58Z" node scripts/send-subscription-receipts`

@jbuck R?

### DIFF
--- a/scripts/send-subscription-receipts.js
+++ b/scripts/send-subscription-receipts.js
@@ -1,0 +1,68 @@
+'use strict';
+
+require('habitat').load();
+
+var stripe = require('stripe')(process.env.STRIPE_API_KEY);
+var hatchet = require('hatchet');
+var async = require('async');
+var moment = require('moment');
+
+var stripe_charge_list_opts = {
+  created: {
+    lte: moment.unix(new Date(process.env.BEFORE_DATE)).valueOf()
+  },
+  limit: 100
+};
+
+var charges;
+
+function process_charge_data(charge, done) {
+  if (!charge.invoice) {
+    return done();
+  }
+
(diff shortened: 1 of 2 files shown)


scripts/send-subscription-receipts.js
```diff
@@ -0,0 +1,73 @@
...
+require('habitat').load();
+
+var stripe = require('stripe')(process.env.STRIPE_SECRET_KEY);
+var hatchet = require('hatchet');
+var async = require('async');
+var moment = require('moment');
+
+var stripe_charge_list_opts = {
+  created: {
+    lte: moment.unix(new Date(process.env.BEFORE_DATE)).valueOf()
```
jbuck: This needs to be converted to a regular Unix epoch count in seconds, not the Javascript variant that counts in ms. You could remove the moment dependency and change it to:

`new Date(process.env.BEFORE_DATE).valueOf() / 1000`

jbuck:
@cadecairos this works, but walking all 51,892 charges, 100 charges at a time is going to be slow. I think we should modify mofo-transaction-storage to store where a transaction is recurring or not so we can do this more quickly...

jbuck:
@cadecairos `select id from stripe where recurring = TRUE and timestamp <= '2015-11-28T03:50:58Z';`

jbuck:
@cadecairos so basically we can just add the list of ids as an array to this script, and run that through

cadecairos:
Lovely!

cadecairos:
oh wow, I was working on my master branch! derp

PR #1055, opened 2015-12-04
fixing one pixel in padding

ScottDowne:


ScottDowne:
Not sure when we broke this, but the other amount border is missing a single pixel on the bottom border.

r? @alicoding

alicoding:
r+

PR #1058, opened 2015-12-04
Add HTTP tests

jbuck:


jbuck:
@alicoding r?

alicoding:
R+ after travis go green

PR #1077, opened 2015-12-08
Fix #1076 Revisit the privacy policy checkbox copy

### LINKED ISSUE #1076: Revisit the privacy policy checkbox copy
I have reported the same issue on Bedrock before ([Bug 937986](https://bugzilla.mozilla.org/show_bug.cgi?id=937986), https://github.com/mozilla/bedrock/pull/1448). The current copy on the privacy policy checkbox reads "I’m okay with **you** handling this info as **you** explain in your privacy policy", while the error message reads "To submit this form **you** must agree with our privacy policy". The former is confusing as **you** means Mozilla so it should be revised.

Found in Mozlando.

kyoshino:


locales/en-US/messages.properties
```diff
@@ -30,14 +30,13 @@ give_to_mozilla=Give to Mozilla Today
...
 sign_up_for_email=Sign up for email updates
-i_am_okay_with=I'm okay with you handling this info as you explain in your <a href="https://www.mozilla.org/en-US/privacy/">privacy policy</a>.
 dont_miss_important_news=Don't miss important news about the work your donation makes possible.
 cvc_info=You can find your three-digit CVC number on the back of your credit card
 credit_card_expiration_month=Credit Card Expiration Month
 credit_card_expiration_year=Credit Card Expiration Year
 email_info=We&rsquo;ll email you a receipt for your donation. We won&rsquo;t send you other email without your permission.
 pp_acknowledge=To submit this form you must agree with our privacy policy
-privacy_policy=I&rsquo;m okay with you handling this info as you explain in your <a href="https://www.mozilla.org/privacy/" target="_blank">privacy policy</a>
+privacy_policy=I’m okay with Mozilla handling my info as explained in <a href="https://www.mozilla.org/privacy/" target="_blank">this Privacy Policy</a>
```
jbuck: @alicoding is it important to use the HTML entity instead of the right quote character?

jbuck:
@alicoding r?

PR #1081, opened 2015-12-09
Allow inserting Optimize.ly images

kyoshino:
At MozLando, we are working on an [Engagement sprint](https://wiki.mozilla.org/Engagement/Mozlando_Projects/Pitch_Presentations) to boost email opt-ins on the Thank You page. I’d like to specify the background image quickly via Optimizely but the CSP doesn’t allows us to do so. Can we make this change? It should not be harmful. Thanks!

PR #1087, opened 2015-12-10
Allow changing newsletter type in HTML

kyoshino:
As part of the Mozlando Engagement project mentioned in #1081, we’d like to dynamically change the newsletter slug which is currently hardcoded in a route. It can be a hidden input control like https://www.mozilla.org/en-US/newsletter/

@alicoding r?


kyoshino:
@jbuck Can you please push this to production?

kyoshino:
Good news: Our test actually begins on Monday. We don't push you guys any more  :laughing:

PR #1103, opened 2015-12-14
Fix #1098 Allow changing newsletter type in HTML

### LINKED ISSUE #1098: Allow changing newsletter type in HTML
@kyoshino filed in this #1087 but the newsletter parameter was not getting sent to the server correctly:

```
--- Hatchet message ---
{ app: undefined,
  event_type: 'send_post_request',
  timestamp: '2015-12-12T00:23:25.448Z',
  data:
   { url: 'https://basket-dev.allizom.org/news/subscribe/',
     json: true,
     form:
      { format: 'html',
        lang: 'en-CA',
        newsletters: undefined,
        trigger_welcome: 'N',
        source_url: 'https://donate.mozilla.org/',
        email: 'jon@jbuckley.ca',
        country: undefined } } }
-----------------------

If you look at the [route payload validation](https://github.com/mozilla/donate.mozilla.org/blob/6e9826315b76355ef512a580e3e3dc8d262fa1bf/server.js#L75-L81) a working patch must add a validation there too, and default to 'mozilla-foundation' for the case where older page code is running.
```

kyoshino:
I’m not sure this works or can be tested locally… @jbuck Please take a look.

### DIFF
--- a/routes/signup.js
+++ b/routes/signup.js
@@ -5,7 +5,7 @@ module.exports = function(transaction, callback) {
   var payload = {
     format: 'html',
     lang: transaction.locale,
-    newsletters: 'mozilla-foundation',
+    newsletters: transaction.newsletters,
     trigger_welcome: 'N',
     source_url: 'https://donate.mozilla.org/',
     email: transaction.email,
--- a/server.js
+++ b/server.js
@@ -82,6 +82,7 @@ module.exports = function(options) {
         validate: {
           payload: {
             locale: Joi.string().min(2).max(12).required(),
+            newsletters: Joi.string().default('mozilla-foundation'),
             email: Joi.string().email().required(),
             country: Joi.string().allow('')
           }
--- a/components/input.jsx
+++ b/components/input.jsx
@@ -16,7 +16,7 @@ module.exports = React.createClass({
   },
   getInitialState: function() {
     return {
-      value: "",
+      value: this.props.value || "",
       valid: true,
       error: ""
     };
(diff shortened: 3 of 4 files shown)


routes/signup.js
```diff
@@ -8,7 +8,7 @@ module.exports = function(transaction, callback) {
     form: {
       format: 'html',
       lang: transaction.locale,
-      newsletters: 'mozilla-foundation',
+      newsletters: transaction.newsletters || 'mozilla-foundation',
```
ScottDowne: I suspect `transaction.newsletters` is always going to be `undefined`.

We need to add the Hidden name to this array: https://github.com/kyoshino/donate.mozilla.org/blob/1098-newsletter-type/components/signup.jsx#L48

ScottDowne: So the submit line should look like `submit={["email", "country", "newsletters"]}`

ScottDowne:
Is the value in `<Hidden>` intended to be updated via optimizely? Not sure how that's going to work, because I think react needs to be told to update.

ScottDowne:
@kyoshino thoughts?

routes/signup.js
jbuck: I think this should be transaction.newsletters only, and we can set a default in the route validation

server.js
```diff
@@ -75,6 +75,7 @@ module.exports = function() {
         validate: {
           payload: {
             locale: Joi.string().min(2).max(12).required(),
+            newsletter: Joi.string().required(),
```
jbuck: Lets add a default of `'mozilla-foundation'` here with https://github.com/hapijs/joi/blob/master/API.md#anydefaultvalue-description

kyoshino:
Thank you @ScottDowne and @jbuck! Um, yes, `submit={["email", "country", "newsletters"]}` in `signup.jsx` was what I was looking for. This hidden value will be updated via Optimizely. I've just modified `input.jsx` instead of adding `hidden-input.jsx`. Does it work?

alicoding:
What happen with this ticket @ScottDowne ?

ScottDowne:
I'm not sure if we this functionality is still needed. Thoughts?

I was delayed because react and optimizely don't work well together, and it's not obvious, which causes things to not work as expected. I also had a hard time with this during the rush of end of year campaign.

ScottDowne:
I'm closing this, it feel off the radar and was tricky to get working, and now has merge conflicts.

If this is functionality we still want, we need to reopen and get it working before Dec, or there likely won't be enough time.

PR #1107, opened 2015-12-14
Fix issue #1101 - zeroDecimal of undefined

### LINKED ISSUE #1101: Uncaught error: Cannot read property 'zeroDecimal' of undefined
```
 » 13 Dec 2015 16:55:59.759 2015-12-13 21:55:59.451045+00:00 app web.1 - - message="Uncaught error: Cannot read property 'zeroDecimal' of undefined" stack="TypeError: Uncaught error: Cannot read property 'zeroDecimal' of undefined at Object.amountModifier.modify (/app/scripts/amount-modifier.js:20:31) at Object.amountModifier.paypal (/app/scripts/amount-modifier.js:36:17) at routes.paypal (/app/routes/index.js:165:33) at Object.exports.execute.internals.prerequisites.internals.handler.callback [as handler] (/app/node_modules/hapi/lib/handler.js:96:36) at /app/node_modules/hapi/lib/handler.js:30:23 at internals.Protect.run.finish [as run] (/app/node_modules/hapi/lib/protect.js:64:5) at exports.execute.finalize (/app/node_modules/hapi/lib/handler.js:24:22) at each (/app/node_modules/hapi/lib/request.js:378:16) at iterate (/app/node_modules/hapi/node_modules/items/lib/index.js:36:13) at done (/app/node_modules/hapi/node_modules/items/lib/index.js:28:25) at request._protect.run (/app/node_modules/hapi/lib/validation.js:59:20) at internals.Any.applyFunctionToChildren.internals.Any._validateWithOptions (/app/node_modules/hapi/node_modules/joi/lib/any.js:657:16) at root.validate (/app/node_modules/hapi/node_modules/joi/lib/index.js:102:23) at Object.internals.input.postValidate [as input] (/app/node_modules/hapi/lib/validation.js:131:20) at exports.payload (/app/node_modules/hapi/lib/validation.js:29:22) at each (/app/node_modules/hapi/lib/request.js:378:16)" tags=error timestring=2015-12-13T21:55:59.443Z Exception Context 
```

There are 6 of them in logentries.

alicoding:
closes #1101

### DIFF
--- a/lib/currency-for.js
+++ b/lib/currency-for.js
@@ -2,8 +2,6 @@ var currencies = require('../data/currencies.js');
 module.exports = {
   stripe: Object.keys(currencies),
   paypal: Object.keys(currencies).filter(function(i) {
-    if (!currencies[i].hasOwnProperty('disabled')) {
-      return i;
-    }
+    return currencies[i].disabled !== 'paypal';
   })
 };
--- a/lib/currency-for.js
+++ b/lib/currency-for.js
@@ -1,0 +1,9 @@
+var currencies = require('../data/currencies.js');
+module.exports = {
+  stripe: Object.keys(currencies),
+  paypal: Object.keys(currencies).filter(function(i) {
+    if (!currencies[i].hasOwnProperty('disabled')) {
+      return i;
+    }
+  })
+};
--- a/server.js
+++ b/server.js
@@ -19,6 +19,7 @@ var routes = require('./routes');
 var goodConfig = {
   reporter: require('good-console-logfmt')
 };
+var currencyFor = require('./lib/currency-for.js');
 
 if (process.env.NPM_CONFIG_PRODUCTION === 'true') {
   goodConfig.events = {
@@ -98,7 +99,7 @@ module.exports = function(options) {
         },
         validate: {
           payload: {
-            currency: Joi.string().min(3).max(3).required(),
+            currency: Joi.any().valid(currencyFor.stripe),
             amount: Joi.number().required(),
             frequency: Joi.string().min(6).max(7).required(),
             stripeToken: [Joi.string().required(), Joi.number().required()],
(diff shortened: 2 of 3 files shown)


alicoding:
@jbuck or @ScottDowne or @cadecairos R?

lib/currency-for.js
```diff
@@ -0,0 +1,9 @@
+var currencies = require('../data/currencies.js');
+module.exports = {
+  stripe: Object.keys(currencies),
+  paypal: Object.keys(currencies).filter(function(i) {
+    if (!currencies[i].hasOwnProperty('disabled')) {
```
jbuck: Could this be simplified to `return !currencies[i].hasOwnProperty('disabled')` ?

jbuck: Also, I think we should check this for `currencies[i].disabled !== 'paypal'`

jbuck: Maybe add the same check for the Stripe list as well?

alicoding: Stripe basically doesn't have that property at all and I believe we want all currencies in the list, so not sure if we need this now.

server.js
```diff
@@ -94,7 +95,7 @@ module.exports = function() {
         },
         validate: {
           payload: {
-            currency: Joi.string().min(3).max(3).required(),
+            currency: Joi.any().valid(currencyFor.stripe),
```
jbuck: Do you need to add `.string()` or `.required()` on each of these lines?

jbuck: Wait, nevermind about `.string()`

alicoding: Not required to do that anymore since if the value doesn't match any of the values in the array then it will fail.

jbuck:
@alicoding this looks great, but I think we should add some http tests here to ensure things are working as expected

server.js
ScottDowne: Did you try it without required?

I remember some weirdness with min(3).max(3) would pass as undefined if I didn't say required.

jbuck:
@alicoding r+ when this goes green

server.js
jbuck: I did a test and got the following response:

```
{ statusCode: 400,
  error: 'Bad Request',
  message: 'child "currency" fails because ["currency" is required]',
  validation: { source: 'payload', keys: [ 'currency' ] } }
```

PR #1109, opened 2015-12-14
remove unique ids from frequency inputs and add test that moves frequ…

ScottDowne:
…ency inputs above amount buttons

### DIFF
--- a/components/donation-frequency.jsx
+++ b/components/donation-frequency.jsx
@@ -53,20 +53,23 @@ module.exports = React.createClass({
   },
   render: function() {
     var frequency = this.state.frequency;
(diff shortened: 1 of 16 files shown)


ScottDowne:
I moved these into two pull requests to try to reduce the size.

The first one is a refactor. It makes it so the unique ids in radio button related components are actually unique, so we can reuse these components.

The second one reuses these components to run a test with moving the frequency input. To test the second commit, the url is http://localhost:3000/en-US/?test=frequency-move-test

@alicoding r?

alicoding:
can you rebase @ScottDowne ?

alicoding:
This one is r+

PR #1111, opened 2015-12-14
Add fs.watch error handling

jbuck:


jbuck:
@ScottDowne r?

ScottDowne:
Little red flag that the catch does nothing, but, I don't think it should stop landing this.

Even with the catch as is, it's better than it was, so R+

If you think we should do something with the catch, can you file a followup?

ScottDowne:
Or, even a comment in the catch would probably be enough? maybe?

PR #1112, opened 2015-12-14
Make https for Facebook and Twitter share URL

yfdyh000:
I don't know whether this project requires a Bugzilla number.


jbuck:
Nah, it's fine to just file/fix issues here. Thanks for the PR!

PR #1113, opened 2015-12-14
Make currency a required key

jbuck:


jbuck:
@ScottDowne r?

server.js
```diff
@@ -120,7 +120,7 @@ module.exports = function(options) {
           schema: {
             id: Joi.string(),
             frequency: Joi.string().valid("monthly", "one-time"),
-            currency: Joi.string().min(3).max(3),
+            currency: Joi.any().valid(currencyFor.stripe).required(),
```
alicoding: Why change this one too?

server.js
```diff
@@ -158,7 +158,7 @@ module.exports = function(options) {
...
```
alicoding: and this

server.js
ScottDowne: I think at this point, this is just covering ourselves to not break what we pass back.

ScottDowne:
@jbuck is there a test for a paypabl disabled currency being used in paypal?

jbuck:
@ScottDowne are you referring to this test? https://github.com/mozilla/donate.mozilla.org/pull/1113/files#diff-0ffce28eb3811edea299bff9abe7ca50R27

jbuck:
@ScottDowne Added the test!

jbuck:
@alicoding yeah, I'm trying to be consistent about our validations; I think it's something we can gradually improve though.

PR #1118, opened 2015-12-15
Exchange rate API

alicoding:


alicoding:
@jbuck R?

tests/http/static-files.test.js
```diff
@@ -99,14 +99,6 @@ describe('static files', () => {
       type: types.html,
       cache: cache.short
     },
-    '/exchange-rates/rates-backup.json': {
```
jbuck: I think you should leave these here until they're actually removed

alicoding:
@jbuck done :)

jbuck:
@alicoding r+ on travis going green after re-addding the test

Issue #1106, opened 2015-12-14
Build and test a short survey for donors


valianttry:
I figured we should create a bug for implementation.

The final copy is linked here:
https://github.com/mozilla/donate.mozilla.org/issues/746

@TylerDowner set up a SurveyGizmo account. @adamlofting suggested we create a simple popup survey. 

@adamlofting feel free to edit labels or Milestone; etc!

adamlofting:
I'll work on this asap.

adamlofting:
After some investigation today, SurveyGizmo's pop-up is not something we'd want to use.

Their pop-up doesn't display the survey, just a (badly styled) message that links out to the full survey page.

Screenshot:
<img width="834" alt="screenshot 2015-12-16 13 09 03" src="https://cloud.githubusercontent.com/assets/1901816/11842978/e57be930-a3fe-11e5-838d-0b44516f8cae.png">

So, in order to improve this, I've added a new page to the donate repo which includes the survey within an iframe, and looks like this:

<img width="758" alt="screenshot 2015-12-16 14 04 35" src="https://cloud.githubusercontent.com/assets/1901816/11843028/20418cd2-a3ff-11e5-81a3-a41bc5149518.png">

I've just opened a pull request for this new page so we can get it onto staging and test it more thoroughly.

adamlofting:
@valianttry what are you thoughts on where we should offer the survey?

I think we have three options, which we can use on either of our thank you pages:

1. Redirect X% of people straight to the survey
2. Promote the survey with a pop-up messge to X% of people
3. Display a message / link within the content of the thank you page for X% of people

**Current thank you pages for reference:**

Initial Thank You page with email opt-in ask:
<img width="847" alt="screenshot 2015-12-16 13 14 12" src="https://cloud.githubusercontent.com/assets/1901816/11843104/9730bfac-a3ff-11e5-8bdb-b4daaaefda2e.png">

Second thank you page with social share ask:
<img width="964" alt="screenshot 2015-12-16 13 14 03" src="https://cloud.githubusercontent.com/assets/1901816/11843108/9dd3cdf4-a3ff-11e5-81f1-c6ee32be73fc.png">

adamlofting:
Here's the link to preview the survey on staging: https://donate.mofostaging.net/en-US/survey/

(this isn't in the donor flow right now so no one is currently seeing this)

adamlofting:
[...]

PR #1124, opened 2015-12-16
Setup a page to host the surveygizmo survey

adamlofting:
Part of #1106 

This adds the survey page so we can look at it on staging, but it doesn't include this page in the donor flow yet.

R? @alicoding or @ScottDowne


ScottDowne:
Looks good. I see an extra whitespace in <header /> and I wonder why make
surveygizmo a separate component.

pages/survey.jsx
```diff
@@ -0,0 +1,25 @@
+import React  from 'react';
+import Footer from '../components/footer.jsx';
+import Header from '../components/header.jsx';
+import SurveyGizmo from '../components/surveygizmo.jsx';
+import { IntlMixin } from 'react-intl';
+
+var ThankYou = React.createClass({
```
alicoding: Just module.exports here?

pages/survey.jsx
```diff
@@ -0,0 +1,25 @@
...
+  render: function() {
+    var className = "row new-flow-test";
+    var language = this.props.locales[0];
+    if (this.props.test) {
+      className += " " + this.props.test;
+    }
+    return (
+      <div className={className}>
+        <Header />
+        <SurveyGizmo language={language}/>
```
alicoding: Where do you use this? The language.

adamlofting:
Good point re seperate component. The first attempt was working with their embed code, but this isn't needed with the simple iframe version. I'll tidy up quickly now.

alicoding:
Just few comments but this looks good to me.

adamlofting:
Thanks both, I'm merging this now to get it on staging.

Issue #1080, opened 2015-12-09
Build version of form with reduced fields in Stripe Checkout


adamlofting:
Test ticket is #1079

ScottDowne:
Just a heads up, with stripe checkout, there is no way to remove address and keep the name field that I can see in the docs.

https://stripe.com/docs/checkout#integration-custom

We can turn on zip code without address, and we should.

valianttry:
yes, zip code without address. Make it so! :rocket:

ScottDowne:
@valianttry you only commented on the zip bit, but didn't say anything about name field also going away.

Does that mean you're fine with name being removed as part of address? We don't have a choice in the matter, but I am just making sure it's not a surprise.

valianttry:
It is perfectly fine by me. When someone opts in to our email list right now, they do not provide their name anyway (only email address). That means we can't personalize emails, e.g. "Dear Katie,"

So name is of no use to us at the moment. I would feel better if @adamlofting agreed with me :+1:

ScottDowne:
@valianttry OK perfect.

ScottDowne:
It's cool, stripe checkout doesn't show the zip field until they have entered a few other fields.

It's like "surprise, you have to type more :D".

valianttry:
We should just be sure to watch our fraud velocity on this form variant to make sure we don't get a spike, or at least one so big it becomes a drag.

ScottDowne:
We might be able to look in Stripe to see how many have failed because of address fields and fraud to get an idea.

ScottDowne:
But should still keep an eye open, because once a fraudster see the hole, they might target it, vs now which they would give up.

ScottDowne:
Wasn't too hard: https://github.com/mozilla/donate.mozilla.org/pull/1115

Should be ready way before the current test finishes :D

jbuck:
@valianttry should I remove the "Dear {name}" customization from the email we send out?

valianttry:
Let's just use "Hello," as our salutation.

jbuck:
@valianttry only downside to using that string is that it'd require translation, and while it was being translated users would be presented with the english string

valianttry:
We can omit a salutation altogether for expediency.

jbuck:
okay, solid

adamlofting:
@ScottDowne is there a way to activate this test without adding the parameter? (Like we've done with class names in Optimizely for the other tests)

[...]

PR #1128, opened 2015-12-16
duplicate of one-page with reduced address fields for a/b test

adamlofting:
This is a dupe of `one-page.jsx` with `billingAddress` set to `false` in the `<SingleForm...>`

R? @jbuck

### DIFF
--- a/pages/one-page-reduced-address.jsx
+++ b/pages/one-page-reduced-address.jsx
@@ -1,0 +1,33 @@
+import React from 'react';
+import Footer from '../components/footer.jsx';
+import Header from '../components/header.jsx';
+import SmallPrint from '../components/small-print.jsx';
+import SingleForm from '../components/single-form.jsx';
+
+var OnePage = React.createClass({
+  mixins: [require('react-intl').IntlMixin],
+  render: function() {
+    var className = "row new-flow-test";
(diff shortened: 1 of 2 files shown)


adamlofting:
Re #1080

pages/one-page-reduced-address.jsx
```diff
@@ -0,0 +1,33 @@
...
+  render: function() {
+    var className = "row new-flow-test";
+    if (this.props.test) {
+      className += " " + this.props.test;
+    }
+    return (
+      <div className={className}>
+        <Header locale={this.props.locales[0]} alt={this.getIntlMessage('donate_to_mozilla')}></Header>
+        <SingleForm
+          billingAddress={false}
```
ScottDowne: We have another patch for this, sorry for the lack of communication.

However, for next time, there is still some query params we would need to remove and ensure the old page passed in true.

You can see the landed patch here: https://github.com/mozilla/donate.mozilla.org/commit/3ca57ba5a1f2d580ca66eb8058cf231cc8f94cfc

Again, sorry about that.

adamlofting:
No worries. I was going to wait to hear from you, but the patch only took two minutes so it was worth the risk :+1:

PR #1130, opened 2015-12-16
Fix caching for local dev

alicoding:


alicoding:
@jbuck R?

jbuck:
@alicoding r+ on travis going green

PR #1131, opened 2015-12-16
Change how exchange rates api works

alicoding:
We moved the directory where we store the exchange rates and how we fetch the file. It is now committed to the repo and we probably want to do this once a day?


alicoding:
@jbuck R?

assets/js/analytics.js
```diff
@@ -87,10 +87,6 @@ module.exports = function() {
   var rates = null;
   if (cc && cc !== 'USD') {
     rates = loadJSON('/api/exchange-rates/latest.json');
```
jbuck: Lets require the exchange rates file here, so it'll be built into analytics.js statically.

jbuck:
@adamlofting I'd like for you to look at this too. The reason I thought we should do it this way is because it'll be more reliable than the existing XHR-based solution and we'll be able to update exchange rates when we want, as opposed to tagging it along with a deploy. I think we should work on a greenkeeper-esque bot to keep the currency rates updated.

jbuck:
@alicoding r+ from me, but lets wait for @adamlofting too

adamlofting:
R+ this works for me.

The exchange rates only need to be approximate for the sake of Optimizely comparisons. 

So long as they're updated weekly or even monthly that will be fine. But let's make sure it doesn't need a human to remember to make the update. We humans are rubbish at that.

PR #1137, opened 2015-12-17
WIP further monthly popup

ScottDowne:


ScottDowne:
@alicoding The css. r?

I also noticed we should run a preemptive validate before we open the popup.

ScottDowne:
If you r+ this can you and jbuck ship it in the morning-ish and loop adam in if he's around to setup the test? I probably won't be around in the morning much.

PR #1140, opened 2015-12-17
Change signup API to be more testable

jbuck:


jbuck:
This will make me more confident in landing #1103

jbuck:
@alicoding r?

alicoding:
r+

PR #1146, opened 2015-12-18
Add more strings for issue 654

alicoding:


alicoding:
@ScottDowne or @flodolo if you can give me r+ for this? :)

edrushka:
@alicoding 
for 2) "tip our hat" is related to the dapper fox's hat. It's an idiomatic English expression used to admire and respect what someone has done.

flodolo:
There might be an equivalent figure of speech in the target language, what I'm asking is to add a note saying that this snippet is associated to such a image (you know it, they don't have the same context).

edrushka:
We can provide the links on staging so they can see: 
    http://scottdowne.github.io/eoy-snippets/2015-spectacular-snippet-1/

    http://scottdowne.github.io/eoy-snippets/2015-spectacular-snippet-2/

    http://scottdowne.github.io/eoy-snippets/2015-spectacular-snippet-3/

alicoding:
@edrushka the copy for staging link doesn't match with what's on google docs though, so not sure that will be helpful unless you are referring to dapper fox only?

ScottDowne:
I can add any copy we want to those staging links, just let me know.

I again had to grab it from the design comps.

alicoding:
@ScottDowne it will be helpful for translators to see the context of the page when doing the translation.

It will be really nice actually if we can plug these pages in donate.mozilla.org repo... what do you think?

ScottDowne:
@alicoding I think it's a good idea, and you're welcome to take that task :P

edrushka:
I'd like to get this up asap so some ideas:
- don't show the screenshots
- explain in the email that the copy doesn't match because we've already translated what you see on staging, and what we need now is the longer text (should mention that the design is responsive)
Let's not block on this because the copy on staging doesn't match.

alicoding:
@flodolo I added more context for each strings.

flodolo:
The comments look good, at this point let's leave out the note about markup. Is @edrushka able to send out the email before we push these strings out?

alicoding:
@flodolo I don't think she gets to that yesterday, but I think I can send an email out to dev-l10n. Is it right that we want to mention in the email that we want these strings by December 22nd but if they can't get them done by then we will use external service to do the translation, but they still get the chance to comment/review the strings right?

flodolo:
I didn't realize the last message for tiles was sent to dev-l10n, dev-l10n-web is a better choice (that's also where Erica's initial message was sent).

Content of the message:
* Apologize for the request coming so late and in the Holiday period :cry: 
* We need these locales done: list
* We need these strings by Dec 22 (end of day, which timezone?). 
* We plan to use them on Dec XX, but need the strings in place by Dec YY (to explain the deadline).
* If you are not able to provide the localization, it's understandable given the deadline but we'll need to use an external agency to get the localization for the missing locales. You'll still have a chance to review the translation if you want (are we sure about that? When would be able to accept comments/reviews?)
* Apologize again, profoundly :sob: 

Note that you'll expose the strings to all locales, good to explain they're not needed, and definitely not urgently.

And then, as soon as this is over, regroup to figure out to avoid deadlines like this again ;-)

alicoding:
@flodolo no, I haven't send anything to dev-l10n, yet. But, I just subscribed to dev-l10n-web and will send the message using Erica's message and ask them there.

PR #1151, opened 2015-12-21
Issue 960 - Currency Symbol

### LINKED ISSUE #960: Add currency symbols to currency switcher
Following the discussion in #911 it might help some users if they can see currency symbols in the currency switcher:

> Another issue might be that the currency switcher isn't obvious to people in es-MX who don't recognise the currency code 'EUR' (we've all got used to this UI now but our audience don't have the same knowledge). Should we test adding the currency symbols back into that drop-down to make it more explicit what it does. 

Filing this as a follow up task.

alicoding:
Closes #960
Added the appropriate currency symbols to the currency selection dropdown.
(via @renvrant)


alicoding:
@adamlofting does this needs to be ab test?

adamlofting:
Discussed on call today: no need to ab test given other priorities. This can land.

Issue #1099, opened 2015-12-13
Paged for 5xx error rates > 1% 2015-12-13T03:26:00


jbuck:

jbuck:
I see one 500 error in the heroku metrics

jbuck:
I think this might be a Hapi bug. I see a lots of requests like:

```
 » 12 Dec 2015 22:25:51.342 2015-12-13 03:25:51.278414+00:00 heroku router - - at=info method=HEAD path="/����%20wwwroot.rar" host=donate-mozilla-org-us-prod.herokuapp.com request_id=bd1395f9-21b7-410f-a1d7-21c1009c03c0 fwd="112.11.81.35,54.239.129.197" dyno=web.1 connect=0ms service=18ms status=500 bytes=333 
[...]

PR #1153, opened 2015-12-21
Return a bad request when we can't complete a redirect

jbuck:


jbuck:
This will fix #1099

jbuck:
@alicoding r?

alicoding:
r+

Issue #981, opened 2015-11-28
On-call page: AWS CloudFront 504 Gateway Timeouts


jbuck:
@ScottDowne was paged today for 6 504 Gateway Timeouts occurring within two minutes of each other. This thread will help us figure out what we can do to fix this problem now, and improve our response in the future.

* The email isn't clear on what the problem is, nor where you can look to figure out what it means
* Scott didn't have access to AWS to see the CloudWatch alarm, and more importantly see that it went away almost immediately
* Scott checked that we were still receiving donations in Logentries, which he really enjoys using
* We couldn't find the failed HTTP request in Logentries
  * But Scott did find another error: https://github.com/mozilla/donate.mozilla.org/issues/980
* I was able to find the requests that errored out in my local copy of the cloudfront access logs:

```
       timestamp        |                path                | status | response_time
------------------------+------------------------------------+--------+---------------
 2015-11-28 16:11:57-05 | /api/paypal-redirect/single/en-US/ |    504 |             1
 2015-11-28 16:11:55-05 | /api/polyfill.js                   |    504 |         30055
 2015-11-28 16:11:45-05 | /api/paypal-redirect/single/en-US/ |    504 |         30014
 2015-11-28 16:10:25-05 | /api/polyfill.js                   |    504 |         30018
 2015-11-28 16:10:16-05 | /api/polyfill.js                   |    504 |         30005
 2015-11-28 16:10:13-05 | /api/polyfill.js                   |    504 |         30013
 2015-11-28 04:16:21-05 | /api/paypal-redirect/single/es/    |    500 |             1
 2015-11-28 04:16:18-05 | /api/paypal-redirect/single/es/    |    500 |             1
 2015-11-28 04:15:33-05 | /api/paypal-redirect/single/es/    |    500 |         25902
```

### Action Items
- [x] We should update the Cloudwatch alert to be more descriptive about what it means and specific steps you can take to investigate issues. Also add this to the local `docs/` folder
- [x] We need everyone on the team to have access to AWS, and directions on where to look when something happens
- [x] We need to make the Cloudfront access logs accessible to everyone else on the team. I wouldn't have been able to diagnose the issue without it
- [x] We need to disable 5xx error caching on the Cloudfront server

PR #1156, opened 2015-12-21
Don't cache 4xx and 5xx errors

jbuck:


jbuck:
Will fix #981 

@alicoding r?

alicoding:
r+

Issue #212, opened 2015-08-11
Look at results of page performance services and file actionable tickets


jbuck:
We have a staging CDN. Hooray! Our performance is kinda disappointing. Boo!

https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fdonate.mofostaging.net%2F&tab=desktop
http://www.webpagetest.org/result/150811_KF_196M/

@adamlofting Can we look at this issue at standup tomorrow and figure out who's on the hook for filing follow-up issues based on these results?

adamlofting:
wohoo :fireworks: 

We have a bug in place to work on this here: #9 

But I didn't want to start tweaking these things until we're happy with our fundamentals. I think we'll be able to work on this in the next heartbeat.

jbuck:
* https://mobiforge.com/design-development/measuring-page-weight
* https://varvy.com/pagespeed/critical-render-path.html
* https://css-tricks.com/prefetching-preloading-prebrowsing/

jbuck:
* http://www.nateberkopec.com/2015/11/05/page-weight-doesnt-matter.html
* http://www.nateberkopec.com/2015/10/21/hacking-head-tags-for-speed-and-profit.html
* http://www.nateberkopec.com/2015/10/07/frontend-performance-chrome-timeline.html
* https://www.igvita.com/2014/05/20/script-injected-async-scripts-considered-harmful/

adamlofting:
I think many of these have been filed. Should we close this now @jbuck ?

PR #1157, opened 2015-12-21
Add preconnect hint for async resources

jbuck:


jbuck:
Will fix #212 

@alicoding r?

alicoding:
r+ after travis going green with adding those slashes to the `<link>` tag :+1:

PR #1171, opened 2015-12-23
Remove two sequential tests

alicoding:
Our server API changed in a way that causes the signup on our old form to break and I don't know if it's worth keeping it up to date and working on this old form. I think it is better to just remove this and maintain our current form.

### DIFF
--- a/tests/selenium/start.js
+++ b/tests/selenium/start.js
@@ -1,14 +1,10 @@
 var testRunner = require('./test-runner');
-var sequentialStripe = require('./sequential-stripe-test.js');
-var sequentialPaypal = require('./sequential-paypal-test.js');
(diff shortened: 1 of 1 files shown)


alicoding:
@ScottDowne R?

jbuck:
@alicoding don't remove the tests without removing the code too :)

I'd leave the decision up to Scott though.

ScottDowne:
Still not in a rush to touch sequential. Running those tests are indirectly
testing components too. Components we do use. Until we have better test
coverage I'm hesitant to remove tests.

alicoding:
@ScottDowne the tests still needs to be removed because it's throwing error `500` from the server causes it to stuck on those sequential page tests. We either have to fix the server to make it work for both old and new form or totally remove them now.

PR #1173, opened 2015-12-23
Fix ordering for listener

alicoding:
In order to make sure our query options such as `frequency`, `amount`, `presets`, and `country` still get fired first before we set up the listener we need to make sure `form.updateState` is being called first.


alicoding:
@jbuck R?

jbuck:
@alicoding r+

PR #1193, opened 2015-12-29
Update copy for spectacular snippet 1

ScottDowne:


locales/en-US/snippet.properties
```diff
@@ -26,7 +26,7 @@ snippet_var_e=Mozilla, the organization behind Firefox, is proudly non-profit, n
...
-snippet_var_f=HELLO there: We know you love Firefox, but did you know a non-profit called Mozilla built Firefox? Thanks to donations, Mozilla does a lot more than build this awesome browser. We teach people to code, fight for online privacy, and protect this amazing thing called the Web for future generations. Just a few times a year Mozilla asks for donations, and usually only a tiny portion of Firefox users give. <highlight>If everyone reading this chipped in {currency}{amount} to help Mozilla, we could wrap this fundraiser up in under an hour and tip our hat to you! <i>Thank you</i>.</highlight>
+snippet_var_f=HELLO there: We know you love Firefox, but did you know a non-profit called Mozilla built Firefox? Thanks to donations, Mozilla does a lot more than build this awesome browser. We teach people to code, fight for online privacy, and protect this amazing thing called the Web for future generations. Just a few times a year Mozilla asks for donations, and usually only a tiny portion of Firefox users give. <em>If everyone reading this chipped in {currency}{amount} to help Mozilla, we could wrap this fundraiser up in under an hour and tip our hat to you!</em> <i>Thank you</i>.
```
jbuck: You could probably update other locales too, right?

ScottDowne: Yup, I was doing this quick as a template as to what needs to be done.

ScottDowne: We need to get this for all en langs, and we need to move change <highlight> and move </highlight> for all locales.

ScottDowne: We could probably leave `<highlight>` as such, but I don't want to have to constantly explain and change it to everyone on the snippet team whenever we try to use it. One source of truth please.

PR #1223, opened 2016-01-05
Fixes issue #600

### LINKED ISSUE #600: TypeError: setting a property that has only a getter
If we are using latest version of Karma we will get this error because it uses a version of socket.io with broken dependency.

Workaround that we are doing currently is in webpack.config.js we are doing this:

```
    new webpack.ContextReplacementPlugin(/buffer/, require('buffer')),
```

and require a known version of buffer 

```
    "buffer": "3.4.3",
```

We should update this once it has been fixed upstream.

alicoding:
Closes #600

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "babel-preset-react": "6.3.13",
     "blankie": "1.1.1",
     "boom": "3.1.1",
-    "buffer": "3.4.3",
     "css-loader": "0.23.1",
     "enhanced-require": "0.5.0-beta6",
     "eslint": "1.10.3",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -52,7 +52,6 @@ module.exports = {
     new webpack.ProvidePlugin({
       'fetch': 'imports?this=>global!exports?global.fetch!whatwg-fetch'
     }),
-    new webpack.ContextReplacementPlugin(/buffer/, require('buffer')),
     new ExtractTextPlugin("style.[hash].css", {
       allChunks: true
     }),


alicoding:
@ScottDowne R?

ScottDowne:
If I understand this, this is the fix that allowed us to remove the above code.

https://github.com/mozilla/donate.mozilla.org/commit/9d103e36906682a56360db26adc3921260ec7d71

And the parent is https://github.com/mozilla/donate.mozilla.org/commit/ebc91f889bfdfb389d978f3258b9d51b9c5752da

In the parent, your above patch should be problematic, in the fix, your patch should be fine.

Correct?

Oddly, when I run npm run test with the parent commit, even without your patch(so with the buffer hack), I get the error.

ScottDowne:
From my tests, the buffer fix never worked.

Obviously this is false, it worked, but right now if I go back and try commits, going all the way back to early Nov, I cannot find a patch with the buffer hack did anything, and removing it seems to be pointless too.

The updates to Karma removes the issue, and removing the seemingly silly buffer hack seems to be safe.

ScottDowne:
I suspect maybe this is what's giving me a hard time: https://github.com/karma-runner/karma/blob/master/package.json#L274

ScottDowne:
An R+ because as far as I can tell, this doesn't break anything, and because I'm going to dig into this more to answer my unanswered question. It's really bugging me, and I don't like unanswered questions.

Facts I can determine using my system, and this is time consuming to nail down.

1. karma "0.13.10" is what introduced these issues initially.
2. the buffer hack doesn't seem to work anymore, as soon as you use "0.13.10" it fails with or without the buffer.
3. somewhere between "0.13.10" and what's on master, this got fixed.

ScottDowne:
The error I keep seeing is. https://github.com/karma-runner/karma/issues/1782

It's unrelated to the above, but it is similar in cause, the error is similar in location (when it happens and the packages involved) and I was confused as to what the original issue actually was, which made me think buffer was initially fixing something in socket.io, when actually buffer was the thing being updated that was causing the issue.

I locked down the fix to socket.io, then I could play with different versions of buffer to understand this.

PR #1225, opened 2016-01-05
add missing strings for ID locale

### LINKED ISSUE #670: id/share page isn't fully translated
I know this is a fairly low priority, so instead of fixing, I filed.

https://donate.mozilla.org/id/share/ <- not translated.

Not sure if that's a bug, or if it's simply missing those translations.

alicoding:
closes #670


alicoding:
@ScottDowne R?

ScottDowne:
R+

PR #1254, opened 2016-01-16
Update German language to fix wrong translation of "your_email"

ftbastler:
"your" translates to "ihre" not "ihree"


jbuck:
@ftbastler Thanks!

PR #1358, opened 2016-03-01
New German Thunderbird Donate strings

koehlermichael:


koehlermichael:
Never mind, seems like I overlooked it on Pontoon. I tried to localize from there.

PR #1363, opened 2016-03-03
Styling the TB Thank You

vazquez:
TB-specific Share page has been created and styled, but the Share buttons have old copy.


components/footer.jsx
```diff
@@ -4,6 +4,10 @@ import { IntlMixin, FormattedHTMLMessage } from 'react-intl';
 var Footer = React.createClass({
   mixins: [IntlMixin],
   render: function() {
+    var firefox_footer = this.getIntlMessage('firefox_footer');
+    if (window.location.pathname.indexOf('thunderbird') !== -1) {
```
ScottDowne: Can this be a prop we pass in? Kinda like how we override the message in checkbox now?

vazquez: This is easier because it eliminates the need to go to every thunderbird-specific JSX layout and pass in the {message}

ScottDowne: Ah, I see.

ScottDowne: What if we had seperate footers that may or may not wrap this one, and pass in the message?

We would still need to go into each thunderbird page and switch it to use the new component, but we wouldn't need to pass in the message on each one. Better?

vazquez: We could have a different footer, and apply that component only to the TB-specific layouts. It just seemed like overkill to have a different component just for one string change. That is why we decided to query the pathname.

PR #1364, opened 2016-03-03
Remove IE comment hack

alicoding:


alicoding:
@ScottDowne R?

jbuck:
Is this safe to remove?

alicoding:
@jbuck apparently that never worked or at least on production right now, and we are now using Stripe which handles that part really well.

ScottDowne:
R+

PR #1380, opened 2016-03-08
making a thunderbird and mozilla specific dir in components for proje…

ScottDowne:
…ct specific components


ScottDowne:
@alicoding just sorting out some files a bit in components, it's large, and I was about to add another.

ScottDowne:
@alicoding updated.

alicoding:
r+

PR #1397, opened 2016-03-11
Update react2

alicoding:


alicoding:
@ScottDowne R?

alicoding:
@ScottDowne the tests are finally passing now \o/ Travis is now happy and green :D

tests/selenium/single-page-thunderbird-paypal-test.js
```diff
@@ -17,13 +17,13 @@ module.exports = function(driver, By, done) {
   driver.wait(function() {
     return driver.getCurrentUrl().then(function(url) {
       var result = true;
-      if (url.indexOf('http://localhost:3000/en-US/thunderbird/thank-you/') === -1) {
+      if (url.indexOf('http://localhost:3000/en-CA/thunderbird/thank-you/') === -1) {
```
ScottDowne: purpose of this change?

I'm fine with it, just curious.

tests/lib/langURLParser.tests.js
```diff
@@ -1,47 +0,0 @@
-var langURLParser = require('../../scripts/langURLParser.js');
-var should = require('should');
-
-describe('langURLParser.js', function() {
-  it('should return true if given path /en-US/ in the paths list', function() {
```
ScottDowne: We don't need these anymore?

tests/http/bad-actors.test.js
```diff
@@ -5,11 +5,13 @@ var instance = server({ useDomains: false });
...
+      url: '/����%20wwwroot.rar',
+      headers: {
+        'Accept-Language':"fr,en;q=0.5"
+      }
     }, (response) => {
-      should(response.statusCode).equal(400);
-      should(response.result.message).equal('Location cannot contain or convert into non-ascii characters');
-
+      should(response.statusCode).equal(302);
+      should(response.headers.location).equal('/fr/');
```
ScottDowne: Can you explain the changes in this file?

pages/thunderbird/share.jsx
```diff
@@ -2,13 +2,17 @@ import React  from 'react';
...
 import ThankYouHeader from '../../components/thank-you-header-thunderbird.jsx';
-import { IntlMixin } from 'react-intl';
 
 var ThankYou = React.createClass({
-  mixins: [IntlMixin],
+  contextTypes: {
+    intl: React.PropTypes.object
+  },
+  propTypes: {
+    test: React.PropTypes.string
```
ScottDowne: Is this required?

pages/thank-you.jsx
```diff
@@ -1,27 +1,40 @@
...
+    router: React.PropTypes.object,
+    intl: React.PropTypes.object
+  },
+  propTypes: {
+    country: React.PropTypes.string.isRequired,
+    test: React.PropTypes.string,
+    location: React.PropTypes.object
+  },
+  getInitialState: function() {
+    let query = qs.parse(this.props.location.search.replace("?", ""));
```
ScottDowne: Not a fan of this, feel like it's going backwards in terms of awesome technology.

pages/thank-you.jsx
```diff
@@ -1,27 +1,40 @@
...
   render: function() {
     var className = "row thank-you-page";
-    var locale = this.props.locales[0];
-    var signUpOrSocial = (<Social language={locale}/>);
-    if (this.props.params && /^(en|de)(\b|$)/.test(locale)) {
-      signUpOrSocial = (<Signup country={this.props.country} email={this.props.email} locales={this.props.locales} />);
+    var locale = this.context.intl.locale;
+    var signUpOrSocial = (<Social />);
+
+    if (/^(en|de)(\b|$)/.test(locale)) {
```
ScottDowne: Why remove the check for `this.props.params`?

tests/selenium/single-page-thunderbird-paypal-test.js
alicoding: Our default locale for the app is en-US, so I thought testing it with different locale is not a bad idea

tests/lib/langURLParser.tests.js
alicoding: We do all our route redirect server side now.

tests/http/bad-actors.test.js
alicoding: The route-file-content is handling those 404 now and we don't hit this anymore and its doing a proper redirect now.

pages/thunderbird/share.jsx
alicoding: Good practice for eslint

pages/thank-you.jsx
alicoding: We don't need it anymore

pages/thank-you.jsx
alicoding: Propose of something else?

pages/survey.jsx
```diff
@@ -23,3 +24,5 @@ module.exports = React.createClass({
     );
   }
 });
+
+module.exports = Survey;
```
ScottDowne: `module.exports = React.createClass` is no longer supported?

alicoding: As I state in mofo eng call that doing it this new way is better

tests/selenium/single-page-thunderbird-paypal-test.js
ScottDowne: Agreed

tests/lib/langURLParser.tests.js
ScottDowne: Makes sense.

tests/http/bad-actors.test.js
ScottDowne: I didn't notice this was a test, thus the change looked very specific and hard coded. As a test result, that actually clears things up.

pages/thank-you.jsx
ScottDowne: Good question, I wasn't very clear.

It looks like you're parsing a query string, where as before the query string was parsed for us by react router. I know we changed how we're handling react router quite a bit, so that might be the reason for the change? But, I really liked that part about react router.

https://github.com/mozilla/donate.mozilla.org/blob/master/components/client.jsx#L8

That ^^ I liked that. Can that not work anymore?

pages/thank-you.jsx
ScottDowne: It looks like dead code. Even on master. Is that why we don't need it?

ScottDowne: I followed it from when we first added it, and `this.props.params` has always been there, and never looks like it's doing anything, so I'm not feeling a lot of confidence in removing it until I know what it was doing.

pages/survey.jsx
ScottDowne: agreed

ScottDowne:
Just two comments still concerning me.

There is also a nasty rebase mostly caused by thunderbird, which, I don't mind doing for you if you don't mind me doing your rebase :)

pages/thank-you.jsx
alicoding: The reason we has that before is when someone hit our route with older version of react router `:locale` param can be empty sometimes, but now it's being managed by react-intl, so we don't need to do that check.

pages/thank-you.jsx
alicoding: The problem is how react router handles routes server side and thus we don't have access to that yet unless we parse that and make it an object. The way we used to do it was client side only that's why

ScottDowne:
@alicoding this is what I have after the rebase https://github.com/mozilla/donate.mozilla.org/pull/1444

I feel like it's still a WIP because there are areas that I did quickly, and would like another pass at it.

PR #1407, opened 2016-03-17
WIP mailchimp integration for thunderbird

ScottDowne:


ScottDowne:
@alicoding r? It's working pretty well I maybe...

mixins/form.jsx
```diff
@@ -411,7 +417,13 @@ module.exports = {
         submitting: true
       });
       submitProps = form.buildProps(props);
-      this.submit("/api/signup", submitProps, this.signupSuccess, this.signupError);
+      this.submit(url, submitProps, success, error);
     }
+  },
+  signup: function(validate, props) {
+    this.doSignup("/api/signup", validate, props, this.signupSuccess, this.signupError);
```
alicoding: What do you think if we change this api route to `/api/signup/basket` and the other one `/api/signup/mailchimp` instead? I just don't like the name for `/api/mailchimp` there tbh.

alicoding:
This looks good to me except that one request.

mixins/form.jsx
ScottDowne: My thought process (and it might not be the greatest idea) was making a thunderbird specific path felt weird as it's not really anything specific to thunderbird, hence the name mailchimp, but I also didn't want the main path we use to become obscured behind a specific. So keeping it signup made sense.

It means most cases we won't need to think about  mailchimp or thunderbird, just signup, and can ignore the other.

alicoding: I feel like this can be confusing and that's why I suggested at least naming it with a specific name in the path to make it clear that this signup path is for `basket` and the other `mailchimp`. Right now, if anyone who's not familiar with the repo and looking at the `/api/signup` would not be able to tell. Also, `/api/mailchimp` is an ugly route name in my opinion... I feel like the route name is not descriptive enough.

ScottDowne: What if signup is one path, and we can pass mailchimp/exact target in (default to exact target)?

alicoding: If you can simplify the code to make it work for both then I'm up for that. We can probably pass in the props (options) to say which service provider we want to use.

ScottDowne: I changed it, but a bit different than your initial request or my initial counter proposal.

I see you issues with the path `/api/mailchimp` it's not specific and I have no issues with making it specific. So it's not `/api/signup/mailchimp`

I did not add a `/api/signup/exacttarget` as I still like having a generic signup path that we use for 90% of our signups.

Thoughts?

alicoding: I would be okay if you make /api/signup generic and able to accept signup for both ET and mailchimp or any future services

ScottDowne: So my goal is so common paths are default. So that means never needing to specify that I want to use exact target. Exact target is the default under `/signup`, and not an optional flag to be turned on and off. Using it, you wouldn't even need to know that it is even called exact target, it's just the standard signup service we have chosen.

Mailchimp is the option, either via a variable or a different path (path is my preference).

If the path is shareable, which I explored, we would need to find a way to make this still secure: https://github.com/mozilla/donate.mozilla.org/blob/master/server.js#L93-L99

mailchimp returns different things than exact target. So we need to somehow have a variable set of response schema. I'm not super concerned with mailchimps response schema, but I don't want to remove exact targets schema.

ScottDowne: I would rather `signup/basket` and `signup/mailchimp` than a generic signup at this point. As a compromise.

alicoding: I was actually thinking about a single route that can handle both services and you can maybe pass the data in such a way that won't break how we're currently doing it for `ET`

If you make a call to `/api/signup` from thunderbird which uses MailChimp
```
{
  service: 'mailchimp'
  data: {
    ...
    ...
  }
}
```
If you make a call to `/api/signup` from our regular donate page which uses ExactTarget
```
{
  service: 'basket'
  data: {
    ...
    ...
  }
}
```

That way in your route handler you can simply call the right method to handle the data from there, and for the `schema` since both form is very similar I think we can get away with having the server deal with that or just let the client pass everything we need there. Though `Joi` got a method that can deal with schemas differently here as well.

ScottDowne: I don't want to go digging in joi, I'm looking for something simpler. I'll take your first suggestion `/api/signup/basket` and the other one `/api/signup/mailchimp` if it gets an r+, or what I currently have up for review.

If we want to go further, I'm going to say we push it to another ticket as fiddling with joi isn't a blocker to shipping thunderbird, which has dragged on waaay too long now.

alicoding: ok. R+ for changing just the route name then :)

ScottDowne: And passing the data in, both basket and mailchimp have the same structure.

It's what is returned, is def not similar. We would need joi to handle two separate objects. You're welcome to figure it out, but I don't think I have the time.

ScottDowne:
@alicoding OK done, merge if it's still an r+

I'm off to bed.

alicoding:
travis is failing...

ScottDowne:
Fixed.

ScottDowne:
At least I think it's fixed, once Travis gets around to finishing we'll know for sure.

alicoding:
Let's merge this, but do you mind adding a test for this in another PR?

PR #1415, opened 2016-03-23
Webhook callback for stripe.disputes.created - which automatically closes them

cadecairos:
This patch adds a webhook callback, protected by a token, that stripe can hit when a dispute is created. The callback will close the dispute automatically.

@alicoding r?

### DIFF
--- a/routes/index.js
+++ b/routes/index.js
@@ -373,7 +373,7 @@ var routes = {
       event.data.object.id,
       function(closeDisputeError, dispute) {
         if (closeDisputeError) {
-          return reply(Boom.badImplementation('An error occurred while closing the dispute', closeDisputeError));
+          return reply(boom.badImplementation('An error occurred while closing the dispute', closeDisputeError));
         }
 
         reply('Dispute closed');
--- a/routes/index.js
+++ b/routes/index.js
@@ -361,6 +361,24 @@ var routes = {
         });
       });
     }
+  },
+  'stripe-dispute': function(request, reply) {
+    var event = request.payload;
+
+    if (event.type !== 'charge.dispute.created') {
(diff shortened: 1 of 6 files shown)


server.js
```diff
@@ -71,6 +71,15 @@ module.exports = function(options) {
     uri: process.env.APPLICATION_URI
   });
 
+  server.register(require("hapi-auth-bearer-token"), function(err) {});
```
alicoding: we don't need to handle any error there? there could be an internal error there right?

server.js
```diff
@@ -71,6 +71,15 @@ module.exports = function(options) {
...
+
+  server.auth.strategy("stripe", "bearer-access-token", {
+    validateFunc: function(token, callback) {
+      // this = request
```
alicoding: do we need `this`?

alicoding:
travis is failing.

server.js
cadecairos: incredibly, ridiculously unlikely, considering the module versions won't change unless manually changed

server.js
cadecairos: nah, tis cruft

routes/index.js
```diff
@@ -361,6 +361,24 @@ var routes = {
...
+
+    if (event.type !== 'charge.dispute.created') {
+      return reply('This hook only processes newly created disputes');
+    }
+
+    stripe.closeDispute(
+      event.data.object.id,
+      function(closeDisputeError, dispute) {
+        if (closeDisputeError) {
+          return reply(Boom.badImplementation('An error occurred while closing the dispute', closeDisputeError));
```
alicoding: lowercase `boom`

alicoding:
r+

PR #1420, opened 2016-03-23
adding share message to thunderbird share page

ScottDowne:


ScottDowne:
@alicoding might be the last code change in donate for thunderbird! r?

alicoding:
R+

Issue #1368, opened 2016-03-04
Thunderbird email signup


ScottDowne:
We need to figure out what needs to happen for this.

From discussing with a few people, I *think* the best way forward is this.

Put thunderbird signups in exact target as a place to store the emails. We don't need to send any emails from exact target, if we don't want to. From here we can figure out what to do with them.

This unblocks us, and we can improve later. We can import then to something else or integrate another mail service or automate the importing process, or something.

cc @edrushka @alicoding @davidascher @vazquez

edrushka:
Yes, this was how I imagined this would work. Can you report back on:
1) How can we mark TB-specific donors in ET to create a special group?
2) Confirm that we can export a list of names/emails from this group.

davidascher:
3) confirm rkent can import said list into mailchimp and email without
mailchimp managing the double opt in.

ScottDowne:
@edrushka @davidascher 
1. I think we can use a newsletter value for this. Right now we use "mozilla-foundation" I'm thinking thunderbird can use one like "mozilla-thunderbird"
2. I can ask this. I suspect we'll be fine, but I'm confirming.
3. @davidascher I'm not sure I follow.  It sounds like thunderbird needs double opt in? If so, yes, exact target is going to handle this, and I can confirm the exported list is only the double opt out confirmed emails.

davidascher:
Yes I'm sure et will handle double optin (isn't that mandated by law?).
What I'm asking is if mailchimp can do a batch import of a CSV file or
whatever, of whether they need to manage the optin themselves. Should be
easy to figure out.

ScottDowne:
@davidascher I'm not sure how optin and import are connected.

Also, I haven't heard back from the moco contact I got from @adamlofting for this, and it's possible it's not the right person currently.

Is there anywhere anyone can point me to who I need to contact to get this setup in exact target?

Once we have that contact, I can actually answer these questions regarding import. Right now my  guess is as good as yours.

davidascher:
I may be paranoid, but I can imagine setups where an email as a service
company won't send an email to a list of email addresses if it doesn't have
logs and evidence that all of those folks have opted in.

So to answer this question I think just check out mailchimp docs to see if
there is an "import email list" feature.

Make sense?

edrushka:
As for setting this up in ET, have you connected with Stephanie yet? She can probably help you get the ear of MoCo's Salesforce/ET manager.

adamlofting:
[...]

PR #1432, opened 2016-03-28
add a flag to the metadata for thunderbird donations

cadecairos:
For #1368

### DIFF
--- a/routes/index.js
+++ b/routes/index.js
@@ -76,7 +76,7 @@ var routes = {
     };
     var request_id = request.headers['x-request-id'];
 
-    if (transaction.description === "Thunderbird") {
+    if (transaction.description.indexOf("Thunderbird") >= 0 ) {
       metadata.thunderbird = true;
     }
 
--- a/routes/index.js
+++ b/routes/index.js
@@ -76,6 +76,10 @@ var routes = {
     };
     var request_id = request.headers['x-request-id'];
 
+    if (transaction.description === "Thunderbird") {
+      metadata.thunderbird = true;
+    }
+
     stripe.customer({
       metadata,
       email: transaction.email,


routes/index.js
```diff
@@ -76,6 +76,10 @@ var routes = {
     };
     var request_id = request.headers['x-request-id'];
 
+    if (transaction.description === "Thunderbird") {
+      metadata.thunderbird = true;
+    }
```
ScottDowne: Is there a reason you need it as a flag and not a desc? Or you just feel this is better overall?

ScottDowne: accounting relies on this param being description and not a flag. I'm not sure how that process works, though. Feels busted to me.

cadecairos: Does accounting take into account monthly subscriptions, where charges don't have a description set?

This method that uses metadata ensures that everytime a charge comes through, this flag is on the customer object, even for monthly subs

ScottDowne: But does it need to be a boolean?

cadecairos: No, but description remains untouched as "Thunderbird". This flag is for the donation email rendering code

ScottDowne: OK, that works for me. I wonder why we didn't try that in the first place for description?

So this makes me wonder if accounting knows how to get thunderbird subscriptions at all? My guess is no.

Maybe we're not passing description at all for subscriptions and we'll need to pass it into metadata anyway. Thus, we would end up with description and the boolean, which, if you're fine with, so I am?

ScottDowne:
@cadecairos yeah, see this line https://github.com/mozilla/donate.mozilla.org/blob/master/mixins/form.jsx#L311

That changes the description.

PR #1433, opened 2016-03-28
ensure donations go to the proper thank you page

ScottDowne:


ScottDowne:
@alicoding I noticed the the thunderbird donate wasn't redirecting to the proper thank you.

I wasn't willing to do paypal, which also seems to be busted. I'll tackle that tomorrow.

server.js
```diff
@@ -237,7 +238,7 @@ module.exports = function(options) {
       }
     }, {
       method: 'GET',
-      path: '/api/paypal-redirect/{frequency}/{locale}/',
+      path: '/api/paypal-redirect/{frequency}/{locale}/{appName}/',
```
ScottDowne: For some reason `{appName?}` didn't work as optional, so I ensure we pass in "thunderbird" or "mozillafoundation" so that something is always here and not optional. I think in the long run this is better though as it means if we ever need a fourth param, we don't need to trip over an optional third param.

alicoding:
r+

PR #1436, opened 2016-03-29
Webhook for updating recurring charges

cadecairos:
When a charge succeeds, it will hit this endpoint, which will update the charge with it's subscription's metadata and a relevant description

### DIFF
--- a/routes/index.js
+++ b/routes/index.js
@@ -415,6 +415,55 @@ var routes = {
         reply('Dispute closed');
       }
     );
+  },
+  'stripe-charge-succeeded': function(request, reply) {
+    var event = request.payload;
+    var charge = event.data.object;
+
+    if (event.type !== 'charge.succeeded') {
+      return reply('This hook only processes charge succeeded events');
+    }
+
+    stripe.retrieveCharge(
+      charge.id,
+      function(fetchChargeErr, charge) {
+        if (fetchChargeErr) {
+          return reply(boom.badImplementation('An error occurred while fetching the invoice for this charge', fetchChargeErr));
+        }
+
+        if (!charge.invoice || !charge.invoice.subscription) {
+          return reply('Charge not part of a subscription');
+        }
+
+        stripe.retrieveSubscription(
+          charge.invoice.customer,
+          charge.invoice.subscription,
+          function(retrieveSubscriptionErr, subscription) {
+            if (retrieveSubscriptionErr) {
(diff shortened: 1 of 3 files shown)


routes/stripe.js
```diff
@@ -73,5 +73,16 @@ module.exports = {
...
+  updateCharge: function(chargeId, updateData, callback) {
+    stripe.charges.update(chargeId, updateData, callback);
+  },
+  retrieveCharge: function(chargeId, callback) {
+    stripe.charges.retrieve(chargeId, {
+      expand: ["invoice"]
+    }, callback);
+  },
+  retrieveSubscription: function(customerId, subscriptionId, callback) {
+    stripe.customers.retrieveSubscription(customerId, subscriptionId, callback);
```
ScottDowne: I'm fine with or without these wrappers around stripe.etc

server.js
```diff
@@ -282,6 +282,13 @@ module.exports = function(options) {
         auth: "stripe"
       },
       handler: routes['stripe-dispute']
+    }, {
+      method: "POST",
+      path: "/stripe/charge-succeeded",
```
ScottDowne: I wonder if it's wise to put webhook routes behind something like `webhook/stripe/charge-succeeded/dispute-callback/etc`?

routes/stripe.js
cadecairos: I'd prefer not to do it, but I don't want to re-instantiate stripe in the index.js file. I suppose I could just export the stripe object here.

routes/index.js
```diff
@@ -415,6 +415,55 @@ var routes = {
...
+  },
+  'stripe-charge-succeeded': function(request, reply) {
+    var event = request.payload;
+    var charge = event.data.object;
+
+    if (event.type !== 'charge.succeeded') {
+      return reply('This hook only processes charge succeeded events');
+    }
+
+    stripe.retrieveCharge(
```
cadecairos: I make this call to expand the invoice object, which then gives me the subscription id, which I can request.

server.js
cadecairos: The other webhook doesn't follow this convention, so I think we'd want to handle that change in a new issue if we think it's warranted

routes/stripe.js
ScottDowne: Ah, I see.

Hm. I say let's just leave it as you have it for right now. Not worth worrying about it right now imo.

PR #1444, opened 2016-04-01
Update react2

ScottDowne:


pages/thank-you.jsx
```diff
@@ -1,27 +1,40 @@
...
+    router: React.PropTypes.object,
+    intl: React.PropTypes.object
+  },
+  propTypes: {
+    country: React.PropTypes.string.isRequired,
+    test: React.PropTypes.string,
+    location: React.PropTypes.object
+  },
+  getInitialState: function() {
+    let query = qs.parse(this.props.location.search.replace("?", ""));
```
ScottDowne: @alicoding I'm still not fully understanding this change, but I did some console.logs and this.props.email is already what we need. If you look in route-file-contents there is this function `makeFile(renderProps, query)` and that query param does have the correct email value from query string, which is properly passed along in queryParser.js

ScottDowne: so `this.props.email` is correct already. We should need to refetch it.

alicoding: I might have update the queryParser after working on this... If you can assure that it works for the full flow and get the email correctly once donated then feel free to update this code.

ScottDowne: Cool, thanks for the timely response. I'm working on it now, and I wanted to make sure I wasn't missing something obvious.

Now that thunderbird has landed, I noticed it was missing a bunch of updated to really similar pages. So I'm also making the repetitive bits of the thank you page a shared thing. It'll be a good pattern for other pages that are shared.

ScottDowne: Is there a line of code somewhere where we pass in `location`? It's no longer needed and I want to make sure I'm not passing in unused values.

alicoding: The location is in contextType

ScottDowne: So we don't need to remove the `location: pathname` lines in route-file-contents? That's still required/used?

ScottDowne: I tried removing it, and bad things happened, so I suspect they might be unrelated. And I don't need to remove anything else, just  double checking though.

ScottDowne:
@alicoding hm, so I did a log inside the start of makefile in route-file-content.js

I expected the log to only happen the first time I load the page, because of cache right?

However it logged in makefile on each refresh of the page. So I'm not sure how the caching of this works, or maybe it's broken/turned off on local?

alicoding:
The server would always be hit no matter what here, but the caching would definitely help with the loading here which is going to be enabled on our prod only.

ScottDowne:
Oh so caching is not done in the repo but instead at the cloud front end?

alicoding:
yep @ScottDowne

ScottDowne:
Works for me.

alicoding:
r+ from me and lets heavily test this on staging. It will also be a good idea for @cadecairos to do a sanity check for our backend (logging etc) to see if this doesn't effect our performance of the page.

cadecairos:
"+1,412 −2,114" wow.

ScottDowne:
@cadecairos @alicoding I also want someone to confirm that the cloudfront caching of these urls is setup and working. Also that the cache is reasonably to clear.

cadecairos:
> Also that the cache is reasonably to clear.

Can you clarify what you mean here

alicoding:
We have memory issue here :(

@cadecairos How do we look for what might have cause this?

ScottDowne:
@cadecairos we're now building the html dynamically as requested instead of building it server side once.

In order for this dynamic build of files to not be bad, we need to have cloudfront cache it so it's built once and from then on in cloudfront holds it in memory and just serves up the previous page based on the url. I want to make sure this cache doesn't get stuck for 10 years or some nonsense when we push up a change.

PR #1446, opened 2016-04-04
fix bug where changing frequency messes up values

ScottDowne:


ScottDowne:
@alicoding r?

STR:

1. load up the normal page https://donate.mozilla.org/en-US/
2. enter 20 in the other amount
3. hit monthly
4. hit pay with paypal

Expected: the paypal amount should be 20
Actual: it's set to 10, downgrading the donation.

ScottDowne:
This is a bug, when frequency  changes, it changes the presets, and when presets change, they check the value and if it's included in the presets, we update it.

However, if the value in other also is equal to a preset, we don't consider that, and STILL downgrade it, cept, the UI still think it's an other amount, so we never update the input box.

ScottDowne:
Looks like it's failing tests.

ScottDowne:
tests should pass this time.

alicoding:
r+ if travis is happy... Good catch though.

ScottDowne:
Turns out this would only happen via interaction with the form, not from query  string values. So anyone coming from snippet is likely to not have this downgrade. Which means it's a bit fringe.

PR #1447, opened 2016-04-04
update site url for og tag if thunderbird

ScottDowne:


ScottDowne:
@alicoding r?

I did a test, I don't think this is going to solve our issue, but, should be fixed anyway.

And a slim chance my test wasn't great and this does fix it, but I doubt it.

alicoding:
r+

PR #1454, opened 2016-04-06
Update throng

alicoding:


start.js
```diff
@@ -20,7 +20,6 @@ function start() {
 }
 
 // lol @ Infinity actually being used for something meaningful
-throng(start, {
-  workers,
-  lifetime: Infinity
-});
+throng({
```
cadecairos: the first param doesn't need to be an object if it just specifies the worker count

assets/exchange-rates/latest.json
```diff
@@ -1,179 +1,179 @@
 {
```
cadecairos: remove these changes plz

cadecairos:
R+

PR #1459, opened 2016-04-07
revert react 2.0 for now

ScottDowne:


ScottDowne:
@alicoding r?

I needed to merge in reverts for the top 4 commits. the two throng commits, the polyfill service commit, and react 2.0.

Webpack isn't going to easily work for running our code on node. It'll require either an understanding of a rarely used undocumented thing, or good documentation for webpack, both of which I don't have access to.

ScottDowne:
I have other ideas, but I want to talk to you about them before I start because it's going to be a significant change.

ScottDowne:
1. we ditch the precompile :(
2. we do a whole babel cli build of the all the source files. src/->public/ this means we would have a src/components/ and a public/components we then feed the contents of public into webpack, and no longer expect webpack to do any babel/react compile, as it's already done. Right now we kinda do babel twice anyway.

PR #1468, opened 2016-04-08
Remove stuff

ScottDowne:
List of removed pages and their screenshots @adamlofting 

https://donate.mozilla.org/en-US/one-page-monthly/
monthly

https://donate.mozilla.org/en-US/sequential/
sequential

https://donate.mozilla.org/en-US/partial-page/
This is a page that didn't ask the user for a full address
partial

https://donate.mozilla.org/en-US/survey/
survey

https://donate.mozilla.org/en-US/directory-tiles/
directory-tiles jpg


ScottDowne:
@alicoding r?

alicoding:
You'll need to update the test as well.

ScottDowne:
hm, tests passed for me locally, but I may have made changes after the test.

I'm in the middle of something so I'll update in a bit.

ScottDowne:
@alicoding Updated.

Turns out it  was passing locally because old files lived in public. I cleared it locally and it passed.

A side effect I didn't consider when doing the supported_locales stuff, and doing a clear of public like you wanted to on build would have caught this locally. I wonder though if a clear public npm script that runs either manually, or on npm install and npm test would be the best place for this?

alicoding:
R+

PR #1482, opened 2016-04-13
React update

ScottDowne:


ScottDowne:
@alicoding Okay, I think this be done.

I'm willing to chat about some of these changes. Some changes might take you unexpectedly.

alicoding:
I think this looks good to me, nothing is wrong with what you have from the code side and would really like to see your next patch that how small it can be ;)

PR #1522, opened 2016-04-29
remove compression to see if it fixes range issues with something som…

ScottDowne:
…ething aws


ScottDowne:
@cadecairos r?

cadecairos:
R+, if you don't care whether or not to make it based on a config setting, so we can toggle it on and off without worrying about backing it out if other patches need to merge.

PR #1553, opened 2016-05-09
maintain query params whatever they may be for all redirects

ScottDowne:


ScottDowne:
@alicoding assuming this passes travis, r?

This adds and ensure that query params don't get lost during redirects. It's actually pretty simple.

things that should now work:

"localhost:3000/?currency=gbp" -> "localhost:3000/en-US/?currency=gpb"

This one may have worked before the recent react update, or maybe it didn't, but either way, it probably should be fixed.

"localhost:3000/thank-you/?email=scott@emails.everyone" -> "localhost:3000/en-US/thank-you/?email=scott@emails.everyone"

I don't think this ever worked.

"localhost:3000/en-UB/?currency=gpb" -> "localhost:3000/en-US/?currency=gpb"

This fixes the busted locale.

"localhost:3000/en-US/nesigbus/?currency=gpb" -> "localhost:3000/en-US/?currency=gpb"

This fixes the busted path.

"localhost:3000/en-UB/nesigbus/?currency=gpb" -> "localhost:3000/en-US/?currency=gpb"

This fixes the busted path and locale.

All of these should retain their currency/email query.

ScottDowne:
I figure I might need to redo some of this in the near future, but I would rather be feature complete before I rewrite the next chunk.

alicoding:
code looks good to me. R+

PR #1558, opened 2016-05-10
ensure redirects with no query also works, for a bit it was sending i…

ScottDowne:
…n undefined


ScottDowne:
@alicoding r+'ed this in person.

PR #1559, opened 2016-05-10
Redirect tests

ScottDowne:


components/client.jsx
```diff
@@ -26,14 +26,9 @@ function createElement(Component, props) {
...
 function onEnter(nextState, replaceState) {
-  var pathname = langURLParser(nextState.location);
-  var query = nextState.location.query;
-  if (pathname) {
-    if (query.redirect) {
-      pathname += query.query || "";
-      query = {};
-    }
-    replaceState({}, pathname, query);
+  var result = redirectURLParser(nextState.location);
```
ScottDowne: I moved the above functionality into langURLParser to make the functionality easily testable.

I also changed the name of the function as it's doing more than just locale redirects, but generic redirects like `en-US/nseobius` and redirects with query unrelated to the locale.

scripts/redirectURLParser.js
```diff
@@ -1,17 +1,19 @@
...
+    redirectQuery = query.redirect;
   }
 
   // We don't need to redirect.
-  if (!redirectQuery && state.pathname !== "/") {
-    return;
+  if (!redirectQuery && location.pathname !== "/") {
+    return {
+      pathname: ""
+    };
```
ScottDowne: returning a more predictable value in general from this function, and trying to keep the return type the same. This is also better for testing.

scripts/redirectURLParser.js
```diff
@@ -1,17 +1,19 @@
...
-function langURLParser(state) {
+function redirectURLParser(location) {
   var redirectQuery = false;
-  var queryString = state.query;
+  var query = location.query;
 
-  if (queryString) {
-    redirectQuery = queryString.redirect;
+  if (query) {
+    redirectQuery = query.redirect;
```
ScottDowne: All the above changes are name changes.

scripts/redirectURLParser.js
```diff
@@ -29,10 +31,16 @@ function langURLParser(state) {
 
   // invalid lang but with a valid path.
   if (!langTest && Object.keys(pages).indexOf(pathname) !== -1) {
-    return "/" + validLang + pathname;
+    pathname = "/" + validLang + pathname;
+  } else {
+    pathname = "/" + validLang + "/";
   }
-  // default
-  return "/" + validLang + "/";
```
ScottDowne: We are returning once at the end of the function now to reproduce less code related to query string values now being moved into this function and returned to make testing easier.

scripts/redirectURLParser.js
```diff
@@ -29,10 +31,16 @@ function langURLParser(state) {
...
+
+  if (redirectQuery) {
+    pathname += query.query || "";
+    query = {};
+  }
+  return {pathname, query};
```
ScottDowne: This is the bits that used to be in client.jsx `onEnter` now moved in here as explained earlier, easier to test.

tests/http/redirects.test.js
```diff
@@ -0,0 +1,57 @@
+var server = require('../../server');
```
ScottDowne: This are just some simple tests on the server for the redirects from a 404.

ScottDowne: It's only have the picture though, as this is passed off to react router and tested separatly.

tests/lib/langURLParser.tests.js
```diff
@@ -1,40 +0,0 @@
-var langURLParser = require('../../scripts/langURLParser.js');
```
ScottDowne: removed via name change, not sure why github isn't figuring this one out.

tests/lib/redirectURLParser.tests.js
```diff
@@ -0,0 +1,161 @@
+var redirectURLParser = require('../../scripts/redirectURLParser.js');
+var should = require('should');
+
+var tests = [
```
ScottDowne: I made the results much more predictable so now the test can be much more dry.

ScottDowne:
@alicoding r? I commented in the ticket to help explain things.

alicoding:
R+

PR #1562, opened 2016-05-11
ensure multiple query values in 404 redirects get parsed properly, so…

ScottDowne:
… we encode/decode it while it travels through our system

### DIFF
--- a/scripts/redirectURLParser.js
+++ b/scripts/redirectURLParser.js
@@ -6,7 +6,7 @@ function redirectURLParser(location) {
   var query = location.query;
 
   if (query) {
-    redirectQuery = query.redirect;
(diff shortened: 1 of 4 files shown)


ScottDowne:
@alicoding I noticed when I passed the query over, it was parsing it incorrectly and stripping off all but the first query value.

Example "https://donate-mozilla-org-us-staging.herokuapp.com/en-NO/?currency=gbp&amount=10" redirects to "/en-US/?currency=gbp" with amount=10 removed.

Fixing this is easy with encode/decode(URIComponent)

Tests are not really needed for passing in two values because I have added encoding to the tests, and they new fail if we no longer decode.

r?

alicoding:
r+

PR #1563, opened 2016-05-12
reduce work in polyfill by doing server side lang detection

ScottDowne:


ScottDowne:
@cadecairos this should fix it if my hypothesis is correct. R?

Basically, we use server side lang detection now for polyfill.

ScottDowne:
Shoot, issue, it's `*only*` using lang headers server side, it ignores the url path, which it should use first.

server.js
```diff
@@ -244,7 +249,12 @@ module.exports = function(options) {
       method: 'GET',
       path: '/api/polyfill.js',
       handler: function(request, reply) {
-        var features = request.query.features;
+
+        var header = Parser.parse(request.headers["accept-language"]);
```
cadecairos: what happens if accept-language isn't defined?

ScottDowne:
@cadecairos should be ready for review, and I'll want to test this for memory usage, if it fixes it, somehow.

cadecairos:
R+

PR #1576, opened 2016-05-18
 Thunderbird redirects aren't working for subpages like /thunderbird/about/ 

mkmelin:
It was interpreting thunderbird as an invalid locale...


mkmelin:
@edrushka Could we move ahead with this? Or who should I ping?

edrushka:
cc @ScottDowne

scripts/redirectURLParser.js
```diff
@@ -30,7 +30,9 @@ function redirectURLParser(location) {
   }
 
   // invalid lang but with a valid path.
-  if (!langTest && Object.keys(pages).indexOf(pathname) !== -1) {
+  if (!langTest && Object.keys(pages).indexOf(redirectQuery) !== -1) {
```
ScottDowne: Interesting!

I think the result from `i18n.urlOverrideLang(redirectQuery);` is incorrect, and should maybe be fixed internally.

I took a look at the internals of that function, and found this line:

`var pathname = localPath.split('/')[2] || localeCode;`

Seems wrong to me. It's assuming anything with with two levels for the path means the first one must be a locale, and that's not true anymore. (at one point we just had only one level deep for all our pages.)

I would prefer if `urlOverrideLang` was fixed to handle multiple levels int he path. Thoughts?

mkmelin: Yeah, I've just made the suggested changes.

locales/i18n.js
```diff
@@ -19,11 +19,20 @@ function formatLocale(lang) {
 function urlOverrideLang(path) {
   var localPath = path || location.pathname;
   var localeCode = localPath.split('/')[1];
-  var pathname = localPath.split('/')[2] || localeCode;
+  localeCode = (/^[a-z]{2,3}(-[A-Z]{2})?$/.test(localeCode)) ? localeCode : null;
```
ScottDowne: Are you familiar with node and npm?

If so we have a little module that's a list of known langs we can check against.

https://www.npmjs.com/package/langmap

You can see an example of me using this in another site here: https://github.com/mozilla/advocacy.mozilla.org/commit/cbb5421190d29b9396f39fe84e32eaa2d6bb86c8#diff-bc8b2010218e61c4795f2d4928be206bR3

mkmelin: I'm familiar yes. But, this (i18n.js) is client side code, so we would just have to make a copy of that file, not just require it. 
Anyway, seems a bit heavy to me for a minor use like this when we don't really care about the details of the locale.

ScottDowne: Cool cool, works for me.

I've been meaning to move some of this to the server side and grab the locale from the headers, but havn't had a chance. At that point, we can reconsider the viability of a langmap.

ScottDowne:
I like your approach, I'll just run it and see if I can break it, if not, I think we'll be good to merge.

ScottDowne:
I'm going to ship this to production early tomorrow eastern time.

It should be available for testing on staging soon: https://donate-mozilla-org-us-staging.herokuapp.com/

PR #1603, opened 2016-07-19
update to react 15

ScottDowne:


ScottDowne:
@alicoding Seems to work out of the box.

The Selenium tests are busted, but unrelated to this. You'll need to test paypal manually, and disable the paypal tests to test using selenium.

alicoding:
There's no breaking changes from 14 to 15 in our use case, so r+

PR #1604, opened 2016-07-20
some non breaking package updates and removing packages I don't see b…

ScottDowne:
…eing used anymore


ScottDowne:
@alicoding I did a few small updates for non breaking versions. Tested too, seems to be fine. I'm not super worried.

I also removed some packages that I didn't see in the repo being used. However, I'm not sure what they were ever being used for, so I wouldn't mind if you could take a look at the packages I removed and confirm they are not used auto magically somewhere.

package.json
```diff
@@ -46,16 +46,13 @@
     "babel": "6.3.26",
     "babel-core": "6.3.26",
     "babel-loader": "6.2.1",
-    "babel-preset-es2015": "6.3.13",
```
alicoding: This is being used.

package.json
```diff
@@ -46,16 +46,13 @@
...
-    "babel-preset-react": "6.3.13",
```
alicoding: and this

package.json
```diff
@@ -46,16 +46,13 @@
...
     "bestlang": "0.0.1",
     "blankie": "1.1.1",
     "boom": "3.1.1",
     "css-loader": "0.23.1",
     "enhanced-require": "0.5.0-beta6",
     "eslint": "1.10.3",
     "eslint-loader": "1.1.1",
-    "eslint-plugin-react": "3.14.0",
```
alicoding: and this

package.json
alicoding: by `.babelrc`

package.json
alicoding: `babelrc`

package.json
alicoding: `.eslintrc`

package.json
```diff
@@ -108,7 +105,6 @@
     "karma-webpack": "1.7.0",
     "mocha": "2.3.4",
     "moment": "2.12.0",
-    "react-test-context": "0.1.0",
```
alicoding: I don't know about this one, so might be okay to remove.....

alicoding:
Revert that 3 back and r+

ScottDowne:
Updated, merging.

package.json
ScottDowne: I think we were using it in the test framework before there were some react-intl updates.

alicoding: Right! The router and intl stub

PR #1606, opened 2016-08-02
updating react router

ScottDowne:
Primary change, updating react router from 1.0.0 to 2.0.0

I needed to change other things because of a result of react router changes.

I can list them, if you want, but it should be obvious. I can also answer specific questions for changes, which ever is easier.


ScottDowne:
@alicoding ping? r?

components/client.jsx
```diff
@@ -27,13 +26,18 @@ function createElement(Component, props) {
...
   var result = redirectURLParser(nextState.location);
-  if (result.pathname) {
-    replaceState({}, result.pathname, result.query);
+  if (!result.pathname) {
+    return;
+  }
+  if (result.query) {
+    replaceState({pathname: result.pathname, query: result.query});
+  } else {
+    replaceState(result.pathname);
```
ScottDowne: replaceState has a new function signature now. It won't accept pathname with a built in query string if it's passed in as an object, like in line 33, so now I need to make the distinction between the two.

components/client.jsx
```diff
@@ -27,13 +26,18 @@ function createElement(Component, props) {
 
 function onEnter(nextState, replaceState) {
   var result = redirectURLParser(nextState.location);
-  if (result.pathname) {
-    replaceState({}, result.pathname, result.query);
+  if (!result.pathname) {
```
ScottDowne: Just an early return

components/routes.jsx
```diff
@@ -2,27 +2,30 @@
...
-            <Route {...routeOBJ}/>
-          );
-        });
-      })
+function buildPagesFromData(data) {
+  return Object.keys(data).map(function(pageKey) {
+    if (pageKey === "/") {
+      return (<IndexRoute key={pageKey} component={data[pageKey]}/>);
     }
+    return (<Route key={pageKey} path={pageKey.slice(1, -1)} component={data[pageKey]}/>);
```
ScottDowne: Split is removing the opening and closing slashes on the urls, just for react router.

scripts/redirectURLParser.js
```diff
@@ -32,10 +32,11 @@ function redirectURLParser(location) {
   } else {
     pathname = "/" + validLang + "/";
   }
-
   if (redirectQuery) {
+    // Query is encoded as a string,
+    // so we don't parse it, and add it to the pathname
     pathname += decodeURIComponent(query.query || "");
-    query = {};
+    return {pathname};
```
ScottDowne: Instead of an empty object, we'll clean this up and return query: undefined.

This way we don't need to look at an empty object.

scripts/route-file-content.js
```diff
@@ -58,25 +60,30 @@ function routeFileContent(outputPath, callback) {
       );
     }
 
-    var index = React.createFactory(require('../pages/index.jsx'));
```
ScottDowne: Most of this was changed because their API has changed, but I basically copied the render bits from advocacy.mozilla.org, because it was also cleaner.

tests/lib/redirectURLParser.tests.js
```diff
@@ -48,8 +48,7 @@ var tests = [
       }
     },
     result: {
-      pathname: '/de/',
-      query: {}
```
ScottDowne: All these tests are changed because the API is better now, and returns undefined instead of an empty object.

alicoding:
I know you left a lot of comments, but it'd be nice if you write in the description the purpose of this update. Will get to it as soon as I can, but expect a delay review since I'm out on vacation and probably will be busy once I get back.

alicoding:
This looks good to me and I don't think I have any question since I have familiarized myself with React-router 2.0 before and everything makes sense, so R+

PR #1607, opened 2016-08-11
Src dir

ScottDowne:
Hey @alicoding or @cadecairos or @gideonthomas whoever has time and the will to look at this.

I added some comments to explain changes that are not obvious or pattern like.

The general description of these changes are this:

1. remove babel compiling from webpack and remove babel import resolving from webpack.
3. move babel compiling into npm scripts and simplified npm scripts.
4. create a src/dist directory structure for compiled source files.
  6. moved all files that need to be compiled from wherever they were, into src.
  7. updated all paths that were effected by this new directory change. (this is the bulk of the changes)
5. separate build for tests because it doesn't make sense to compile test files for a prod like build.


.eslintignore
```diff
@@ -1,3 +1,4 @@
 node_modules/*
 public/*
-assets/*
\ No newline at end of file
+assets/*
+dist/*
```
ScottDowne: We don't bother linting compiled files.

data/pages.js
```diff
@@ -1,14 +0,0 @@
...
-  '/share/': require('../pages/share.jsx'),
-  '/give-bitcoin/': require('../pages/give-bitcoin.jsx'),
-  '/paypal-donate/': require('../pages/paypal-donate.jsx'),
-  '/thunderbird/': require('../pages/thunderbird/one-page.jsx'),
-  '/thunderbird/about/': require('../pages/thunderbird/about.jsx'),
-  '/thunderbird/thank-you/': require('../pages/thunderbird/thank-you.jsx'),
-  '/thunderbird/share/': require('../pages/thunderbird/share.jsx')
-};
-
-module.exports = pages;
```
ScottDowne: Moved this file into src/data

karma.conf.js
```diff
@@ -20,7 +20,11 @@ module.exports = function(config) {
     webpack: {
       devtool: 'inline-source-map',
       module: {
-        loaders: webpackConfig.module.loaders,
```
ScottDowne: Removing this, as the loaders used in tests and builds are no longer the same.

scripts/build.js
```diff
@@ -1,17 +0,0 @@
...
-
-if (env.get('NPM_CONFIG_PRODUCTION')) {
-  buildCommand = 'start:prod';
-}
-
-var npm = spawn('npm', ['run', buildCommand], { stdio: 'inherit' });
-
-npm.on('close', function(code) {
-  process.exit(code);
-});
```
ScottDowne: We now run the repo like advocacy, which ditched this method of prod/dev. A refresher is this `npmm start` is simply a dev only thing, that does what prod does + watch.

src/components/mozilla/footer.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import Footer from '../footer.jsx';
+import Footer from '../footer.js';
```
ScottDowne: These file names are changed, because technically this is being run as dist, after it's compiled, so technically it's running js and not jsx.

src/lib/analytics.js
```diff
@@ -67,14 +67,14 @@ module.exports = function() {
   }
 
   // 2 DP
-  function toTwoDP (i) {
+  function toTwoDP(i) {
```
ScottDowne: Rest of this file is just lint fixes. Not sure why it wasn't being caught before, but now it is.

src/lib/route-file-content.js
```diff
@@ -60,8 +60,9 @@ function routeFileContent(location, callback) {
       );
     }
 
-    var Index = require('../pages/index.jsx');
-    FS.makeTree(Path.join(__dirname, '..', 'public', location)).then(function() {
+    var Index = require('../pages/index.js');
+    var dirPath = Path.join(__dirname, '..', '..', 'public', location);
```
ScottDowne: I needed to change the path because things are now one more dir in, so I needed to `..` an extra step. I initially only did this path change only in the `nameOfTheFile` below, and missed it on the `FS.makeTree` line. I had to do extra debugging to catch this, so for the future, I now define the path in one spot as `dirPath` so I only need to manage the path in one spot. :)

alicoding:
This LGTM, but I would like to request one thing from you. Can you also document or update the documentation on how to write a component in this repo? Since it is different than what we used to and it may be a bit confusing since there's nothing explaining the process anywhere in the documentation and for new contributor or people who might be looking at the code to help may find it difficult to follow here.

ScottDowne:
@alicoding sounds good, updated the docs.

PR #1616, opened 2016-09-19
WIP to remove pre compile

ScottDowne:


ScottDowne:
@alicoding maybe time to start looking at this, if possible.

@alanmoo I might need to lean on you a bit for review, if it goes to next week.

package.json
```diff
@@ -9,8 +9,8 @@
     "build:exchangerates": "node scripts/exchangerates.js",
     "build:babel": "babel src --compact=true -s --out-dir dist",
     "build:webpack": "webpack --optimize-minimize --optimize-dedupe",
-    "watch:js": "npm-run-all --parallel watch:webpack watch:babel",
-    "watch:babel": "npm run build:babel -- -w -q",
+    "watch:js": "npm-run-all --parallel watch:babel watch:webpack",
```
alicoding: does the sequence in this matter?

tests/http/bad-actors.test.js
```diff
@@ -1,16 +0,0 @@
-var server = require('../../server');
```
alicoding: why this test was removed?

alicoding:
Just one redflag about that test you've removed since that was a bug that we may hit because of how hapijs handle their routing and stuff...

package.json
ScottDowne: Not really, but webpack listens for changes in babel, and webpack does an initial build, along with babel. So it just felt logical to put them in their logical order.

Also with this order, it more likely to reduce log outputs that and not needed, but that's only on average.

tests/http/bad-actors.test.js
ScottDowne: I talked to Jbuck, and he created this test to test that urls with non ascii characters sent us to a 400. However, we now redirect to the home for urls that are not in react-router, so I saw no more need to test that we were doing something we are not.

I couldn't figure out what else this test was trying to do beyond that, and there wasn't anything else in the test saying why a url couldn't contain non ascii characters, so I actually consider it a pretty toxic test.

alicoding: Please do a `curl` test locally to see if it redirects properly and if so then this is r+

ScottDowne: I'm happy to re add this if you can give more info about the bug. Right now, there isn't anything in the test about why it's there.

Although the test would need to be re made, as it's not working like it used to (and hasn't for a while)

ScottDowne:
This is going to staging now.

PR #1619, opened 2016-09-22
Moving some imports from the top of the file to inside a function and…

ScottDowne:
… called using require, so garbage collection can do the thing.


cadecairos:
R+

PR #1624, opened 2016-10-19
Setup logic for handling micro payment switching

ScottDowne:


ScottDowne:
@cadecairos ping! :rocket: 

If you don't think you can find the time today, let me know, so I can make other plans. It's not mandatory that you do this, but if you cannot, I need to know.

ScottDowne:
@cadecairos ping pong :ping_pong:

ScottDowne:
@cadecairos I rebased this, so we can now use this as part 2 of the patch.

cadecairos:
Besides my only comment above, this looks solid and works as expected. R+ once the comment above is addressed.

ScottDowne:
@cadecairos updated, mind taking another look?

ScottDowne:
@cadecairos can wait until tomorrow.

PR #1625, opened 2016-10-20
Fixing warning in donation-props.tests.jsx

ScottDowne:


ScottDowne:
@cadecairos r?

There was a warning in our `npm run tests`, and it looked like it was in a generic exports. So I started to rewrite out exports to have names until I found it.

I ended up finding it in the test `donation-props.tests.jsx` Once I found it, I left the exports changes, as it's better to do it that way. I see no need to undo what I already did, but I also won't go further after finding the issue.

The actual issue in the test was a broken context type being emulated around react router. The values were there, but no longer being plugged in since the API changed and we did a react-router update.

I updated it to the new API, and things seem to be passing without warning.

ScottDowne:
@cadecairos This is just tests, and not on the critical path, so I'm just going to merge.

PR #1626, opened 2016-10-24
Fixing privacy policy locale links.

ScottDowne:
@TheoChevalier just updating what was obvious in locale codes hard coded into privacy links, via find and replace. Mind taking a quick look?

However, I think I still need to create a `privacy_policy_var_c` so the locales know to update.


TheoChevalier:
@ScottDowne Mozilla.org privacy policy link look definitely good. 

For Paypal, they do not seem to redirect automatically. For instance with my browser in French, I’m not redirected to https://www.paypal.com/fr/webapps/mpp/ua/privacy-full 
Maybe that’s because they are using geolocation and I’m in the UK. I don’t know.
 Anyway, using https://www.paypal.com/webapps/mpp/ua/privacy-full doesn’t hurt even if it does not redirect, so I think we’re good.
Thanks Scott!

ScottDowne:
@TheoChevalier yeah, I'm not redirected either. I not 100% sure what kind of silly logic they have, but I'm pretty sure it's a combination of geolocation and cookies, because if I go here: https://www.paypal.com/webapps/mpp/ua/legalhub-full then select a language, I now start seeing pages translated, if I go to a new tab, it's in English again.

I think the pages with a locale code in them is to override the cookie or geolocation versions. So I think the algorithm first checks the url, if a code is in there, it uses that. Then it checks to see if there is a cookie, if it find one, it uses that, then finally, it probably uses geolocation.

ScottDowne:
edit: "if I go to a private tab, it's in English again."

PR #1627, opened 2016-10-24
Update privacy policy link and update en-CA and en-AU.

ScottDowne:
@TheoChevalier review?

This is fixing:

- Updating the privacy policy so it can be translated.
- Removing out of date .properties files in en-CA and en-AU so en-US is used instead.


TheoChevalier:
Looks good, thanks!

PR #1629, opened 2016-10-28
Rename some IN based locale codes.

ScottDowne:


ScottDowne:
@TheoChevalier for the most part, everything works. Review?

I found an issue with gu and gu-IN, which was actually an issue in the langmap package. I made a pull request for that here. https://github.com/mozilla/language-mapping-list/pull/20 So once that lands, we can update the version we use of langmap, and gu should then be enabled.

I see no reason to wait for the langmap patch to land, in order to land this.

ping me if you have any questions.

TheoChevalier:
LGTM, thanks!

PR #1636, opened 2016-11-03
Fixing locale pick for es-MX

ScottDowne:


ScottDowne:
@TheoChevalier thoughts?

ScottDowne:
This is the es messages used in es-MX, and in general, I'm adding a function that we can scale up if we find more cases like this.

TheoChevalier:
I tested the PR on a local instance, and it works as I would expect it to work with different locale codes as accepted-language (mainly tried with es, es-MX, es-CL, and mixing them), so looks good to me 👍

ScottDowne:
@TheoChevalier great. I'll push it up to prod tomorrow morning.

Issue #1640, opened 2016-11-07
Russian Form shows USD vs. Rubles


valianttry:
![donate_mozilla_org___ _mozilla_](https://cloud.githubusercontent.com/assets/9830371/20044402/876febb6-a44d-11e6-962d-52329e109c54.png)

ScottDowne:
This is fixed, should be live in a few minutes.

Issue #1641, opened 2016-11-07
Czech form shows USD instead of Czech koruna


valianttry:
![donate_mozilla_org___darujte_mozille_jeste_dnes](https://cloud.githubusercontent.com/assets/9830371/20044427/c01b1026-a44d-11e6-9b1e-d809321ff7ca.png)

ScottDowne:
This is fixed, should be live in a few minutes.

PR #1642, opened 2016-11-07
Adding russian and czech currency for their locales.

ScottDowne:


ScottDowne:
@TheoChevalier if you don't mind taking a quick look. It's a pretty simple fix. Fixes #1640 and #1641.

TheoChevalier:
Looks good to me, but what’s preventing us from doing that for more locales? (ja: JYP, for instance, and probably many others)

ScottDowne:
Nothing. We just need to sit down and go over all the locales and see
what's missing.

PR #1644, opened 2016-11-09
Default currency for additional locales

TheoChevalier:
@ScottDowne added currencies for Argentina & Chile (Stripe only) + assigned a default currency for several locales. Does that look good?
I’m not sure how to test the localeCountryData array, however. What is it used for?

Also, alpha sorted the arrays for improved readability


ScottDowne:
I'm pretty sure localeCountryData is dead code. We used to have a form that had a country dropdown, and it would be pre selected based on locale.

I havn't had the time to confirm and remove it, but I was planning on doing that soon.

src/lib/get-messages.js
```diff
@@ -2,7 +2,7 @@ import assign from 'object-assign';
 import locales from '../../public/locales.json';
 
 function getMessages(locale) {
-  if (locale === 'es-MX') {
+  if (['es-AR', 'es-CL', 'es-MX'].indexOf(locale) >= 0) {
```
ScottDowne: Should we add es-AR and es-CL empty locale folders with a `.keep` file to the repo to expose these locales?

ScottDowne:
Looks pretty good to me. One comment about a thing I think we can easily add.

src/lib/get-messages.js
TheoChevalier: Yes, I actually did it, but forgot to git add :(

TheoChevalier:
Should be better, good catch :)

PR #1646, opened 2016-11-09
Send donation information into an SQS queue for basket to consume

cadecairos:
This PR borrows functionality in [hatchet](https://github.com/jbuck/hatchet/) so that we can send donation information into an SQS queue for consumption by [basket](https://github.com/mozmar/basket/).

I had to pull out the hatchet functionality because it can't support more than one queue, and we already integrate with sawmill for email signups.

To test, you can confirm that donation data is dumped to console when no BASKET_* env vars are provided. If you'd like to test with an actual Queue, let me know and I'll get you what you need.


ScottDowne:
Hm, why is it that hatchet cannot "support more than one queue"

ScottDowne:
Wouldn't it make a bit more sense to update basket to support multiple queues?

cadecairos:
> Hm, why is it that hatchet cannot "support more than one queue"

[Just the way it's coded](https://github.com/jbuck/hatchet/blob/3ca1f90b1b04e8d1165107d0bf6fa09a99c2aff9/index.js#L41)

> And is a queue defined by where it goes? I'm not sure what you my by "one"

not sure what you're asking here

> Wouldn't it make a bit more sense to update basket to support multiple queues?

We could, however, [we'd still only be able to support queues in one AWS region](https://github.com/jbuck/hatchet/blob/3ca1f90b1b04e8d1165107d0bf6fa09a99c2aff9/index.js#L4), and in this case, we need to use a queue in a different region that the hatchet queue.

routes/index.js
```diff
@@ -167,6 +188,22 @@ var routes = {
...
+              basketData = {
+                last_name: customer.cards.data[0].name,
+                email: customer.email,
+                donation_amount: charge.amount,
+                currency: charge.currency,
+                created: charge.created,
+                recurring: false
+              };
+
+              if (ZERO_DECIMAL_CURRENCIES.indexOf(charge.currency) === -1) {
```
ScottDowne: I wonder if you can put this logic into a function/requirable file.

That way, on line 205, you can just do this:

```
basket.queue({
                last_name: customer.cards.data[0].name,
                email: customer.email,
                donation_amount: fixAmount(charge.amount, charge.currency),
                currency: charge.currency,
                created: charge.created,
                recurring: false
});
```

ScottDowne: Let me know if you think that's better or not. I'm personally not sure.

cadecairos: Yes, this is a good idea

routes/index.js
```diff
@@ -399,10 +440,24 @@ var routes = {
           }
 
           request.log(['paypal', 'checkout', frequency], log_details);
-
+
```
ScottDowne: space

ScottDowne:
Just one tiny nit, but r+

I think just so long as we're all still on the page about how this works from a version 1 point of view, we can probably move this forward.

@cadecairos Thanks for helping with this.

routes/index.js
ScottDowne: spaaaaaaace

cadecairos:
@ScottDowne I update the patch so that the "appName" attribut is passed to basket as "project". It will be either "thunderbird" or "mozillafoundation"

PR #1647, opened 2016-11-11
Fix Italian form on all the pages

TheoChevalier:
Fixes https://github.com/MozillaFoundation/Advocacy/issues/373

### DIFF
--- a/less/components/currency-dropdown.less
+++ b/less/components/currency-dropdown.less
@@ -3,6 +3,10 @@
   text-transform: uppercase;
   text-align: center;
   padding: 0 20px;
+  margin-left: 5px;
+  display: inline-block;
+  width: 120px;
+  vertical-align: top;
   option {
     padding: 0;
   }
--- a/less/components/submit-button.less
+++ b/less/components/submit-button.less
@@ -12,10 +12,9 @@
     vertical-align: middle;
     cursor: pointer;
     border: 1px solid transparent;
-    white-space: nowrap;
     line-height: 1.42857;
     line-weight: 800;
-    padding: 3px 0 7px;
+    padding: 3px 5px 7px;
     -moz-user-select: none;
     color: #fff;
     margin-top: 5px;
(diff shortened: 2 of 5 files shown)


less/components/currency-dropdown.less
```diff
@@ -3,6 +3,10 @@
   text-transform: uppercase;
   text-align: center;
   padding: 0 20px;
+  position: absolute;
+  width: 125px;
+  right: 5px;
+  top: 0;
```
ScottDowne: Oh wow, now that I see this, the currency dropdown is IN the h3, which it really shouldn't.

We should fix them, and put them side by side.

ScottDowne: Mind giving this a try?:

1. pull `CurrencyDropdown` out of the span
2. put `CurrencyDropdown` beside, not inside, the donate now `h3`
3. add a `donate-now-header` to the `h3`
4. add `display: inline-block; width: 220px;` to `.donate-now-header` 
5. add `display: inline-block; width: 120px; vertical-align: top;` to `.currency-dropdown` 

Thoughts? It means you could remove a float right, an absolute position, make it better on screan readers, and possibly not need to add `main-page`?

TheoChevalier:
@ScottDowne Updated! And thanks, that’s way better indeed. The only “downside” I noticed is that the title does not seem to wrap while the page is loading, but I’m not sure we care?

I’ve added a margin-left on the currency-dropdown, so that if the text takes all the available space, we have a small visual margin.
I also had to increase the h3 width for the /about pages, because the form is somewhat wider by 10px, so the currency dropdown was mot aligned anymore.

ScottDowne:
This looks fine to me.

I did poke around a bit as to why `/about/` was 10px wider, and it looks like it's margin is 5px smaller on the left and right, making for 10px. Weird if you ask me, maybe worth making them the same to reduce complexity in the css. But, I don't think that need to be the problem of this patch.

PR #1648, opened 2016-11-12
Add a warning about string length for tweets

TheoChevalier:
Several locales are over limits, will warn them, but adding this as well


TheoChevalier:
done in #1649

PR #1649, opened 2016-11-14
Removing more unused strings

TheoChevalier:
Removing a bunch of strings that are not used in any template (ran `ack string_id` in src folder, I guess that’s enough?)

I’ve left some strings at the bottom of the file (L91-L104) that are still used in templates, but they don’t seem exposed anywhere. @ScottDowne let me know if it’s safe to remove them as well.

I also reorganized the file, grouping strings per page and sorting them in the same way they are displayed on the page. This will be easier for localizers to understand the context and find the strings for testing. I’ve also added a few comments.

### DIFF
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -1,131 +1,104 @@
+# NOTE: strings for https://donate.mozilla.org. Staging server: https://donate.mofostaging.net
+
+# Donate Bitcoin page
+help_protect_the_web=Help protect the open Web.
+donate_butcoins=Donate Bitcoins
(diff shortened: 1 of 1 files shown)


TheoChevalier:
Oh, and I’ve tested locally as much as I could to check that no string were showing up untranslated

ScottDowne:
I'm having a hard time sorting out which ones are being moved, and which are being removed.

I would need to manually go through line by line, to double check that the things you removed are fine to remove.

Is it easy to create two commits in this pull request, one to remove unused strings, and one to move the remaining strings?

PR #1650, opened 2016-11-14
Ensure we don't add array based test strings.

ScottDowne:


ScottDowne:
@vojtechmzl review?

Just for future usage, we don't want query string based tests to trample each other.

previously, if the same query string for test appeared in the url, we would get the class name "test-name,test-name" which would in effect turn off the test.

So my patch makes it so the class name becomes "test-name test-name" this way, even though the classname is duplicated, the test still runs.

vojtechmzl:
Thanks for the heads up on this, the fix makes sense to me.

PR #1651, opened 2016-11-14
Remove more strings

TheoChevalier:
@ScottDowne yeah, should be easier to read now :)

### DIFF
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -1,70 +1,106 @@
+# NOTE: strings for https://donate.mozilla.org. Staging server: https://donate.mofostaging.net
+
+# Donate Bitcoin page
+help_protect_the_web=Help protect the open Web.
+donate_butcoins=Donate Bitcoins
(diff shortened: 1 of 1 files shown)


TheoChevalier:
By the way, I think this is the last bit remaining before I can enable more locales :) Communication to localizers should go out tomorrow with the first request ready: Social posts.
Would be great if we could enable locales at the same time

ScottDowne:
@TheoChevalier Thanks for this :) :rocket:

TheoChevalier:
@ScottDowne I kept some legacy strings here https://github.com/mozilla/donate.mozilla.org/blob/master/locales/en-US/messages.properties#L93-L106 is it safe to remove them even though they are in js files? They don’t seem to appear anywhere

ScottDowne:
I think we can remove those too, but not sure. I think a new patch for just those, and I can review it tomorrow morning, if that would work for you?

Sorry, I saw your comment on legacy strings in the first patch, but you didn't have it here, so I didn't look again for legacy strings.

TheoChevalier:
No worries, opening a new one ;)

PR #1652, opened 2016-11-14
Last batch of legacy strings

TheoChevalier:


locales/en-US/messages.properties
```diff
@@ -88,19 +88,3 @@ i_donated_to_thunderbird=I donated to @mozthunderbird today to #freetheinbox. Jo
...
-first_name=First name
-last_name=Last name
-address=Address
-city=City
-country=Country
-postal_code=Zip/Postal code
-state_province=State/Province
-
-# donate-button.js
-donate_now_amount=Donate { donationAmount } now
```
ScottDowne: This is used here:

donate_now_amount

Just select an amount after using a non paypal currency. It's also used on our https://donate.mozilla.org/en-US/paypal-donate page, but that might be a little less critical, as I've wanted to redo that page for a bit now.

locales/en-US/messages.properties
```diff
@@ -88,19 +88,3 @@ i_donated_to_thunderbird=I donated to @mozthunderbird today to #freetheinbox. Jo
...
-
-
-# Legacy strings
-first_name=First name
-last_name=Last name
-address=Address
-city=City
-country=Country
-postal_code=Zip/Postal code
-state_province=State/Province
```
ScottDowne: I think these are all dead.

All of our address components are pretty much dead, but we still have a legacy test we've never run that uses country drop down on the signup page: https://donate.mozilla.org/en-US/thank-you/?test=country-signup-test

I want to kill this dead, but havn't had the chance. I think if we pull these strings out, we should  consider to fixup or remove some of the code that uses it.

ScottDowne:
I think this looks good now. Merging.

PR #1665, opened 2016-11-18
Ensure we can check accept lang in a case insensitive way, because sa…

ScottDowne:
…fari returns lower case locale codes.

### DIFF
--- a/server.js
+++ b/server.js
@@ -16,9 +16,12 @@ var Joi = require('joi');
 
 var polyfillio = require('polyfill-service');
 var PolyfillSet = require('./scripts/PolyfillSet.js');
-var Parser = require("accept-language-parser");
-var bestLang = require('bestlang');
-var locales = Object.keys(require('./public/locales.json'));
+var getLocale = require('./dist/lib/get-locale.js');
+
+var locales = {};
+Object.keys(require('./public/locales.json')).forEach(function(locale) {
+  locales[locale.toLowerCase()] = locale;
+});
 
 var exchangeRates = require('./assets/exchange-rates/latest.json');
 var routes = require('./routes');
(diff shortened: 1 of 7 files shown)


ScottDowne:
@TheoChevalier this should fix safari.

It's not as big as it looks, because I tabbed in a fairly large function. You can use ?w=1 in the pull request link to ignore whitespace changes, like this: https://github.com/mozilla/donate.mozilla.org/pull/1665/files?w=1

probably easier to read.

The code of these changes are in server.js, which look like this:

```
var locales = {};
Object.keys(require('./public/locales.json')).forEach(function(locale) {
  locales[locale.toLowerCase()] = locale;
});
```

The rough idea is we keep a map of all locales keyed on lower case, with the upper case values, so when we need to check, we can without parsing.

TheoChevalier:
Not sure if you will be able to reproduce, but here’s what I’ve done (with the current patch applied):

- Switched default language to es-MX in macOS Language & region settings
- Restarted Safari
- opened localhost:3000, it loads localhost:3000/es/ instead of localhost:3000/es-MX/

ScottDowne:
Looks like I have tests failing too.

I cannot reproduce, the way Safari changes locale settings isn't allowed in browser stack.

ScottDowne:
@TheoChevalier I found a way to test, I can reproduce, and when I log the lang header, I get es-xl when set to Mexico in mac regions language.

ScottDowne:
@TheoChevalier I'm trying adding es-XL currency. Thoughts?

TheoChevalier:
Looks like mozilla.org loads es-ES in Safari when accepted-language is set to es-XL (es-MX in OS settings)
With this setup on donate, I’m experiencing the redirect issue “http://localhost:3000/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/es-XL/”

I guess we won’t be able to display es-AR and es-MX to Safari users

TheoChevalier:
@ScottDowne LGTM 👍

PR #1670, opened 2016-11-21
Ensure the currency input is uppercased before comparing with the Zer…

cadecairos:
…o Decimal Currency list


ScottDowne:
R+

ScottDowne:
Correction, R+ once green.

PR #1671, opened 2016-11-21
Changes to 'Thank You' and 'Share' Pages

beccaklam:
New graphics added to the thank you and share page headers, as well as an 'share via email' button on the share page social.

Fixes https://github.com/MozillaFoundation/Advocacy/issues/623


less/components/header.less
```diff
@@ -12,4 +10,22 @@
...
   }
+  #heart-gif {
+    @media screen and (min-width:300px) {
+      position: relative;
+      top:-30px;
+      margin-bottom:-55px;
+      width: 50%;
+    }
+    @media screen and (min-width:599px) {
+    top:-50px;
```
ScottDowne: Should probably add two spaces of indent to this and the below media queries.

ScottDowne:
There are some small things I noted, but in general there are also some whitespace nits.

less/pages/thank-you.less
```diff
@@ -14,21 +14,21 @@
   display: inline;
   .header {
     text-align: center;
-    margin-bottom: 0;
+    margin-bottom: 20px;
     margin-top: 0;
-    padding-top: 22px;
+  max-width:none;
```
ScottDowne: Two spaces worth of indent. And put a space between max-width: and none, so like this: `max-width: none;`

less/shared.less
```diff
@@ -514,7 +515,7 @@ select.parsley-error:focus {
 
 @media screen and (max-width: 480px) {
   .container {
-    padding: 10px;
+    border-radius:0px;
```
ScottDowne: You want to add a space between `border-radius:` and `0;`

Also, general practice in mofo is, `0px;` isn't necessary, and should just be `0;`

ScottDowne:
I need to show you how to setup tests, because we should make it possible to test the new background, and the email box, independently of each other. It's not hard, but easier in person.

I'm going to take another pass to ensure nothing you did happens to break something already, but I don't think I'll find anything ;)

src/components/social.js
```diff
@@ -17,20 +18,27 @@ module.exports = React.createClass({
...
-                <i className="fa fa-twitter fa-5x"></i>
+                <i className="fa fa-twitter fa-2x"></i>
                 <div>twitter</div>
               </a>
             </div>
+            <div className="third" id="email">
+              <div className="social-button"></div>
+              <a href={emailShareURL} target="_blank">
+              <i className="fa fa-envelope fa-2x"></i>
+                <div>email</div>
```
TheoChevalier: I guess the string is used for accessibility, so let’s localize it. It should be safe to reuse this string https://github.com/mozilla/donate.mozilla.org/blob/master/locales/en-US/messages.properties#L75
So something like this should work `<div>{this.context.intl.formatMessage({id: 'email'})}</div>`

src/components/social.js
```diff
@@ -9,29 +9,57 @@ module.exports = React.createClass({
     var locale = this.context.intl.locale;
     var twitterShareURL = 'https://twitter.com/share?url=' + appURL +'/' + locale + '/&text=' + encodeURIComponent(this.context.intl.formatMessage({id: 'i_donated_to_mozilla'}));
     var facebookShareURL = 'https://www.facebook.com/sharer/sharer.php?u=' + appURL + '/' + locale + '/';
+    var emailSubject = '';
+    var emailBody = '';
```
ScottDowne: What's the ETA for these?

Also, these are going need to go into a properties.json file likely, as it currently stands. cc  @TheoChevalier

ScottDowne:
I think this code looks solid. I had one question about l10n.

I also wanted to cc @vojtechmzl on this, because of the proposal to make these tests, I believe is in the plan. Right? That said, how do we track success and failure of changes to the thank you page? Is it just bounce rate? Email opt ins or shares?

vojtechmzl:
Hey @ScottDowne, can we make it so that the test variant with the new header can be triggered with a URL parameter, similarly how the frequency selector test worked? 

There is also another ticket in the advocacy repo around adding ShareProgress buttons in order to enable testing different share messages on social and in email (https://github.com/MozillaFoundation/Advocacy/issues/635). Is that something that can be done for EOY?

Email optins and shares are the two principal metrics to keep track off here and actually, on that note, can we attach new Google Analytics and Optimizely events to the email signup and the share buttons for tracking purposes? We currently only track the donations data. I can start a new issue to track that. I can also take a stab at this, if it would be easier.

src/components/social.js
TheoChevalier: Will this be displayed for all locales, and how can it be tested?

src/components/social.js
```diff
@@ -9,29 +9,57 @@ module.exports = React.createClass({
...
+              <a href={twitterShareURL} target="_blank">
+                <i className="fa fa-twitter fa-2x"></i>
+                <div>twitter</div>
+              </a>
+            </div>
+            <div className="third email-share email-share-test" id="email">
+              <div className="social-button"></div>
+              <a href={emailShareURL} target="_blank">
+              <i className="fa fa-envelope fa-2x"></i>
+                <div>email</div>
```
TheoChevalier: still need to be localized ;)

ScottDowne:
Status on this? Is this waiting for something from me?

vojtechmzl:
Sorry for the confusions @ScottDowne, just trying to clarify where things are at. 

Can we test the new design using URL parameter, similarly to the frequency selector test? In order to effectively evaluate this test, we will also need the GA events attached to the signup and share buttons as per ticket #1676.

I also wasn't sure if implementing the Share Progress buttons was done as part of this PR? The ticket for that is here: https://github.com/MozillaFoundation/Advocacy/issues/635.

Thanks.

Thanks.

ScottDowne:
Share progress is not part of this, and yes, it can be turned on and off with normal means

ScottDowne:
@vojtechmzl around tracking the success or fail of this, can we not just check against bounce rate? If the new page produces less bounce rate, it wins?

src/components/social.js
ScottDowne: It wouldn't hurt to localize this, no?

That said, I think it's acceptable to run the test in en-US initially, so I wouldn't call this a blocker.

src/components/social.js
TheoChevalier: Please use the following string IDs: `share_email_subject` for the subject, `share_email_body` for the body

ScottDowne: @beccaklam The strings landed.

So now if you just changed these two lines to:

```
    var emailSubject = this.context.intl.formatMessage({id: 'share_email_subject'})};
    var emailBody = this.context.intl.formatMessage({id: 'share_email_body'})};
```

It should just work. However, those strings were added in another patch after you started yours. So those changes are not tracked in your branch. So if you add what I pasted above it won't do what's expected until we merge it to master, or I(or someone else) shows you how to rebase.

src/components/social.js
TheoChevalier: My previous comment is hidden, but we should be able to reuse the existing string and do:

`<div>{this.context.intl.formatMessage({id: 'email'})}</div>`

ScottDowne:
This landed in another patch.

I pulled in all the history so everything should be tracked and tagged just like this patch. Closing.

PR #1675, opened 2016-11-28
Update faq link

TheoChevalier:
@ScottDowne We just got Spanish completed minutes ago \o/
We now have 20 locales complete, including our target locales (only Polish missing for faq/ways to give).
Updated the links and removed target=_blank, we should be good to push that over to prod whenever it works for you :)


ScottDowne:
hm, thought I said looks good here.

Looks good.

PR #1678, opened 2016-11-29
Update check link

TheoChevalier:
@ScottDowne gha, just oticed I forgot this one…


TheoChevalier:
Oh wait, wrong page. We should probably add an id to the Ways to give page

ScottDowne:
We can make this better if you want, but this is actually fine with me.

PR #1685, opened 2016-12-01
Initial share progress implementation

ScottDowne:


ScottDowne:
Don't merge yet, needs to be tested with share progress turned on, and it's not on yet.

PR #1691, opened 2016-12-03
Not using react router while share progress is used.

ScottDowne:


ScottDowne:
@cadecairos review? This should fix up our issue with share progress.

What was happening is this. Share progress does it's thing on page load. It looks for specific elements, changes them, and the it's done. It's nothing more than that.

With a fresh page load of `/share/`, it works fine. Once you get redirect there from react router, it doesn't actually do a page load, so share progress is never properly able to load it's buttons.

To test.
1. Go to thank you.
2. Signup with a nonsense email or whatever.
3. Once share page loads, try to click either button.

Expected: it should load a facebook/twitter whatever.
Actual: nothing. You're just clicking on a div.

All I did to fix this was do a hard page load via `window.location`.

PR #1693, opened 2016-12-06
Harden social share button CSS for l10n display.

ScottDowne:


ScottDowne:
@TheoChevalier review?

This is fixing this:

pic

To be this:

good

Path to this page is http://localhost:3000/de/share/?test=social-test-with-email

Also make sure the normal share page is acceptable. http://localhost:3000/de/share/

PR #1694, opened 2016-12-08
Add back generic errors

ScottDowne:
@cadecairos I guess you're the best person for this review?

Seems at some point, probably when we moved from generic Stripe to the Stripe widget, we removed our generic error handler, because Stripe widget handles 99% of error messages.

Looks like we found an error it doesn't catch, albeit a fringe error.

To reproduce, locally, use this credit card number: "378282246310005" and the currency CZK

Expected: an error message of sorts.
Actual: nothing.

I'm just adding back basic functionality, not updating the error message (that would require some l10n and I'll do that next and get theo to review) I just need you for the code bits. Does it actually display an error, and more importantly, doesn't break anything else.


assets/exchange-rates/latest.json
```diff
@@ -1,180 +1,180 @@
 {
   "disclaimer": "Exchange rates provided for informational purposes only and do not constitute financial advice of any kind. Although every attempt is made to ensure quality, no guarantees are made of accuracy, validity, availability, or fitness for any purpose. All usage subject to acceptance of Terms: https://openexchangerates.org/terms/",
   "license": "Data sourced from various providers; resale prohibited; no warranties given of any kind. All usage subject to License Agreement: https://openexchangerates.org/license/",
-  "timestamp": 1480647611,
+  "timestamp": 1481238011,
```
cadecairos: Can we not commit this?

ScottDowne: We can not commit it, but I like to commit it from time to time to update the currency rate conversion so we get slightly better metrics.

When I do this, I usually make a separate commit.

This should be harmless though, right? Are you thinking this is an issue?

cadecairos: I see now that it is indeed a separate commit, so I'm okay with including it

ScottDowne:
Eh, I'm not super worried about it being in the QA checklist. But not against it either. It would be a nice to have. Easy enough to do too.

cadecairos:
Is there another more common failure mode for stripe that we should make sure is being caught?

ScottDowne:
This catches anything that, for reasons that are a mystery to me, Stripe checkout accepts but Stripe API fails.

Right now this is the only one I'm aware of. I'm pretty sure there are more, though. Wouldn't hurt to go through here again: https://stripe.com/docs/testing

I haven't gone through that since we started using Stripe checkout.

ScottDowne:
I think this is also going to catch server errors on the API end.

cadecairos:
using this test card #: (4000000000000341) seems to trigger the same error from the Stripe API (Customer created, charge failed to be made)

cadecairos:
IMO I think checking that we don't silently fail here (which would really confuse a donor) is pretty important

ScottDowne:
It does, but just gives us a `card_error` error too. So it's also pretty generic.

cadecairos:
I don't mind that it's generic right now, it's more important that we display any error to the donor right now. I also think that if we make this part of QA, we won't accidentally remove it in the future.

ScottDowne:
Agreed but I won't be adding QA info until Monday :p

cadecairos:
If you file a new issue and link it here, I'll allow a merge :)

ScottDowne:
@cadecairos I filed the QA stuff here: https://github.com/MozillaFoundation/Advocacy/issues/678

PR #1695, opened 2016-12-09
Disable Amex for select currencies.

ScottDowne:


ScottDowne:
@cadecairos or @vojtechmzl or @TheoChevalier not sure who to review. This is a pull request that hides the amex logo for Stripe when the currency is a currency that Amex doesn't support.

ScottDowne:
@cadecairos A few updates.

PR #1699, opened 2016-12-14
Adding test to move payment buttons above privacy statement

vojtechmzl:
Test is triggered with ?test=payment-move included in the URL.

@cadecairos could you please take a look?

Thanks.

### DIFF
--- a/less/pages/single-form.less
+++ b/less/pages/single-form.less
@@ -4,16 +4,15 @@
 
 
 .payment-move-test {
-	display:none;
+  display:none;
 }
 
 .payment-move {
-	.payment-move-baseline {
-		height:0px;
-		margin-bottom:10px;
-
-	}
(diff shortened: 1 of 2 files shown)


less/pages/single-form.less
```diff
@@ -1,3 +1,19 @@
 .payment-section {
   padding-top: 25px;
 }
+
+
+.payment-move-test {
+	display:none;
```
cadecairos: Can you match the indentation in the beginning of the file?

vojtechmzl:
@cade I just pushed the indentation fix, let me know if there is anything else I can do. 

Thanks for your help!

cade:
That feeling when you're primed and ready to unleash hilarity, only to find the @mention typo has been fixed...

flip

PR #1700, opened 2016-12-14
Adding GA events to landing page 

vojtechmzl:
In order to understand where people drop off on the landing page (i.e. ~90% of snippet traffic), we are hoping to add GA events to specific components to see how people interact with the page. The events can be tested by checking the debugger, where a collect function gets called for every event. Under parameters, ec, ea and el correspond to category, action and label for the event.

@cadecairos Could you please review?

Thanks for your help with this.


vojtechmzl:
Thanks @cadecairos!

PR #1702, opened 2016-12-14
Donate button text 386 small width changes

beccaklam:
@vojtechmzl @cadecairos @TheoChevalier Added changes to previous pull request referencing ticket 368 "Add the word "Donate" to the payment method buttons" so that payment method buttons are not broken on mobile/small screens.


TheoChevalier:
Thanks, looks good to me 👍 . I can only spot that the font for Donate is a bit too bold on Safari, but we can revisit later. I’ll go ahead and push this

PR #1704, opened 2016-12-15
Making /about the top landing page

vojtechmzl:
Hey @cadecairos, one of our tests came through with the results below, indicating that directing traffic to /about improved conversion rates compared to the simple one-page form. I just updated the routes.js file to make /about the main landing page and added a new route for /simple to still be able to access the simple one-page form. 

Please let me know if you have any questions.

Thanks for your help.

image


ScottDowne:
This is really interesting, as this change didn't make a difference last year.

I wonder if this would be worth testing again but just for email related asks?

vojtechmzl:
Good point @ScottDowne. We can set up a test for the next email round, which is I believe sometimes next week. Hopefully we can push this change live asap though, as we focused specifically on the snippet traffic this year and we want to maximize the exposure.

src/routes.js
```diff
@@ -11,8 +11,9 @@ var routes = (
       Object.keys(locales).map(function(locale) {
         return (
           <Route key={locale} path={locale}>
-            <IndexRoute component={require('./pages/one-page.js')}/>
+            <IndexRoute component={require('./pages/about.js')}/>
             <Route path='about' component={require('./pages/about.js')}/>
+            <Route path='simple' component={require('./pages/one-page.js')}/>
```
cadecairos: might make more sense as `one-page`?

vojtechmzl:
Yup @cadecairos, one-page route name makes more sense. Updated

PR #1705, opened 2016-12-16
Fix disclaimers style on /about + Safari style issues

TheoChevalier:
This fixes the disclaimer on /about, mainly backporting changes done on the homepage. Does that look good to you @beccaklam ? I’ve tested smaller screen sizes too, but feel free to double check

Before:
capture d ecran - 2016-12-16 a 14 47 38

After:
capture d ecran - 2016-12-16 a 14 47 51

### DIFF
--- a/less/components/payment-buttons.less
+++ b/less/components/payment-buttons.less
@@ -16,7 +16,7 @@
     font-size: 13px;
   }
   .row.medium-label-size.donate-button {
-    font-weight: 900;
+    font-weight: 600;
     font-size: 16px;
   }
 }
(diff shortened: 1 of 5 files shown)


TheoChevalier:
The second commit fixes the layout on Safari
capture d ecran 2016-12-16 a 15 21 17

TheoChevalier:
Shamefully adding a third commit to this PR, because I found out why we’re having font-weight issues on Safari (we don’t have any downloaded font for `font-weight: 800;` or `font-weight: 900;`, we only have “Open Sans Semibold” set for `font-weight: 600;`)

PR #1708, opened 2016-12-16
Blue button test and less text on button tests

beccaklam:
Created tests for donate.mozilla.org to see if blue donate buttons and less text on the donate buttons will improve conversion on the page.

PR #1709, opened 2016-12-16
Fix no amex on blue buttons

beccaklam:
Fixed the blue buttons where American Express is to be hidden for specific locales.


cadecairos:
This commit has been rolled into #1708

PR #1712, opened 2016-12-20
Fixing spacing around payment buttons

vojtechmzl:
Hey @ScottDowne, here is a fix for the spacing issue. Thanks for flagging it.

I didn't know how else to select the elements, so I added new IDs. Let me know if there is a better way.

Thanks.

### DIFF
--- a/less/pages/single-form.less
+++ b/less/pages/single-form.less
@@ -1,11 +1,3 @@
 .payment-section {
   padding-top: 25px;
 }
-
-#section-heading {
-  margin-bottom:14px;
-}
-
-#privacy-policy p {
-  margin-bottom:0px;
-}
--- a/less/pages/single-form.less
+++ b/less/pages/single-form.less
@@ -1,3 +1,11 @@
 .payment-section {
   padding-top: 25px;
 }
+
+#section-heading {
+  margin-bottom:14px;
+}
+
+#privacy-policy p {
+  margin-bottom:0px;
+}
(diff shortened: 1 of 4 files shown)


less/pages/single-form.less
```diff
@@ -1,3 +1,11 @@
 .payment-section {
   padding-top: 25px;
 }
+
+#section-heading {
+  margin-bottom:14px;
+}
+
+#privacy-policy p {
+  margin-bottom:0px;
```
ScottDowne: I was wrong, I don't think this is needed. Thoughts?

less/pages/single-form.less
```diff
@@ -1,3 +1,11 @@
 .payment-section {
   padding-top: 25px;
 }
+
+#section-heading {
+  margin-bottom:14px;
```
ScottDowne: I think we can accomplish the same thing by removing this line: https://github.com/mozilla/donate.mozilla.org/blob/971cb3793f90b94c9a3b73d8b95f8dfee7e8938d/less/shared.less#L472

ScottDowne: Roughly the same thing.

vojtechmzl: Great find. Just pushed an update with the simplified code @ScottDowne. Thanks!

less/pages/single-form.less
vojtechmzl: Agreed. Removed from the latest push.

PR #1713, opened 2016-12-20
Fixing share button css.

ScottDowne:


ScottDowne:
Share progress sometimes puts the clickable area offset from the button container.

This forces it to sit in the top left corner.

@cadecairos

PR #1714, opened 2016-12-22
Import translations from last year for privacy_policy_var_b

TheoChevalier:
@piotrdrag that would be a new string for Polish, since it was not translated last year


ScottDowne:
One note, when we used to have the old privacy policy, we had a link to the privacy policy in the bottom portion of the page too.

ScottDowne:
R+

PR #1715, opened 2016-12-22
Update Polish (pl) localization of Fundraising

piotrdrag:


TheoChevalier:
Thanks a lot, @piotrdrag ! I will go ahead and merge this to avoid having an English string on production. As always, we can push fixes at anytime if required

PR #1716, opened 2016-12-23
Fix margin with payment buttons for long locales

TheoChevalier:
I’m moving the margin on the buttons themselves, because it doesn’t work when the "SECURE" element is wrapping.

Before:
capture d ecran 2016-12-23 a 17 21 24

After:
capture d ecran 2016-12-23 a 17 21 03

### DIFF
--- a/less/shared.less
+++ b/less/shared.less
@@ -469,11 +469,12 @@ i.fa.hint {
 }
 
 .choose-payment {
+  margin-bottom: 0;
   margin-right: 5px;
 }
 
 #secure-label {
-  margin: 4px 0;
+  margin: 4px 0 20px 0;
   font-size: 15px;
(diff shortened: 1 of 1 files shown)


less/shared.less
```diff
@@ -469,9 +469,14 @@ i.fa.hint {
 }
 
 .choose-payment {
+  margin-bottom: 0;
   margin-right: 5px;
 }
 
+input.payment-type + label {
+  margin-top: 15px;
```
ScottDowne: Maybe put the margin bottom on `#secure-label`?

But yeah, good catch.

TheoChevalier: Ah, indeed. Fixed!

ScottDowne:
Looks good to me.

PR #1717, opened 2016-12-23
Implementing winner from payment button testing

vojtechmzl:
Hey @ScottDowne, the results from the payment button test came through and, as per the chart below, simplified blue buttons won, improving conversion rate by 7.8%. I implemented those changes and also removed old tests from the payment-options.less file. Please let me know if you have any questions or concerns.

image

cc @HPaulJohnson

### DIFF
--- a/src/components/payment-options.js
+++ b/src/components/payment-options.js
@@ -37,6 +37,7 @@ var PayPalButton = React.createClass({
         <div className="row payment-logos paypal-logo">
           <p>&nbsp;</p>
         </div>
+        <div className="row medium-label-size less-text">PayPal</div>
       </span>
     );
   },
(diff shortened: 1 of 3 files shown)


src/components/payment-options.js
```diff
@@ -94,7 +93,6 @@ var StripeButton = React.createClass({
           <div className={className}>
             <p>&nbsp;</p>
           </div>
-          <div className="row medium-label-size less-text">{this.context.intl.formatMessage({id: 'credit_card'})}</div>
```
TheoChevalier: How about we keep this string hidden, instead of removing it? I think it’s the only way for screen readers to understand what the button is about

vojtechmzl:
Good point @TheoChevalier on the screen readers. Just added the buttons back hid them. Thanks!

ScottDowne:
This is breaking thunderbird buttons.

vojtechmzl:
I will look into the Thunderbird issue. Thanks for the flag @ScottDowne

vojtechmzl:
@ScottDowne updated the Thunderbird buttons, so they all look the same.

vojtechmzl:
Thanks for your help with this @ScottDowne.

PR #1720, opened 2016-12-27
Adding UAH

ScottDowne:


ScottDowne:
@TheoChevalier Review? No rush on this for today or anything, just whenever.

Fixes https://github.com/MozillaFoundation/Advocacy/issues/691

TheoChevalier:
Tested, looks good, pretty safe to be merged :)
@ScottDowne unrelated to this patch, but I’ve always been wondering where and how values from localeCountryData were used?

ScottDowne:
We have a component for country drop down that we once used in the thank you page which we are not using.

The component still lives because removing it is a bit more work than I thought. I started earlier in the campaign but got pulled away to other priorities.

I am only keeping the list up to date because if I'm going to have code in the repo, it's going to work, or I'm going to do the work to remove it.

We can pull it out next year :)

TheoChevalier:
totally makes sense, thanks for the explanation :) I wasn’t sure if I missed something

PR #1728, opened 2016-12-28
RTL support

TheoChevalier:
WIP, but even if it gets a r+ soon, let’s hold until January anyway as it’s too risky.

Tried to test "tests" as much as possible (like the thank-you alternative design), and also only tested in Firefox Nightly. Pretty sure I’ll find nonsense in Safari…

On the approach itself, I tried to keep the RTL blocks close to the rules it’s overwriting, so that it’s easier to understand while reading the stylesheet. I’ve considered using LESS variables and things like https://github.com/DevelopmentIL/direction.less but in the end I think html[dir="rtl"] is the easiest to understand and maintain. FWIW, this is also what’s used on mozilla.org https://github.com/mozilla/bedrock/blob/master/media/css/firefox/desktop/common.less#L63
But would love to hear your thoughts on that, though.

Currently, the only RTL locale we support is Hebrew (he), but with this patch we can try to get Arabic, at least, and motivate Hebrew localizers to complete the translation.

### DIFF
--- a/less/shared.less
+++ b/less/shared.less
@@ -164,6 +164,17 @@ input[type="submit"] {
   float: left;
   padding-left: 5px;
   padding-right: 5px;
+  word-wrap: break-word;
+}
+
+html[dir="rtl"] {
+  .full,
+  .half,
+  .third,
+  .two-third,
+  .quarter {
+    float: right;
+  }
 }
 
 .full {
@@ -238,6 +249,15 @@ input[type="submit"] {
   float: right;
 }
 
+html[dir="rtl"] {
+  .left {
+    float: right;
+  }
+  .right {
+    float: left;
+  }
+}
+
 /* = Form Fields ========== */
 
 label {
@@ -246,6 +266,10 @@ label {
 
 input[type="radio"] + label {
   margin-left: 10px;
+}
+html[dir="rtl"] input[type="radio"] + label {
+  margin-left: 0;
+  margin-right: 10px;
 }
 
 input[type="text"],
@@ -263,10 +287,7 @@ input[type="email"] {
 
 input[type="email"] {
   padding-right: 33px;
-}
-input[name="lastname"],
-input[name="zip"] {
-  padding-left: 14px;
+  margin-top: 3px;
 }
 
 input[type="text"]:focus,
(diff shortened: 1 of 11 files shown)


TheoChevalier:
Rebased with stuff from https://github.com/mozilla/donate.mozilla.org/commit/f08344a1d7fbb2b6723b08bdf701e24e6ec6d8fd

ScottDowne:
Looking at this now.

ScottDowne:
If there are any merge conflicts between this and https://github.com/mozilla/donate.mozilla.org/pull/1740

We should priorities this, and we'll fix the conflicts when we land https://github.com/mozilla/donate.mozilla.org/pull/1740

less/shared.less
```diff
@@ -164,6 +164,17 @@ input[type="submit"] {
   float: left;
   padding-left: 5px;
   padding-right: 5px;
+  word-wrap: break-word;
```
ScottDowne: Note to myself, double check this doesn't break anything.

ScottDowne: Yup, don't see how this can break anything.

less/shared.less
```diff
@@ -263,10 +287,7 @@ input[type="email"] {
 
 input[type="email"] {
   padding-right: 33px;
-}
-input[name="lastname"],
-input[name="zip"] {
-  padding-left: 14px;
+  margin-top: 3px;
```
ScottDowne: Purpose of adding 3px margin top to email?

Email input is only used on thank-you page. I don't see how rtl could cause it needing an extra 3 px top.

ScottDowne:
Just one small nit, but otherwise I don't  see how this could break anything currently supported.

If we have any issues, we can pretty safely iterate until we have more rtl languages supported.

Just one question, then I think  we can land this.

less/shared.less
TheoChevalier: Right, not related to RTL, fixing this on Chrome and Safari (Note the cut gradient at the top):
capture d ecran 2017-01-24 a 01 41 20

less/shared.less
TheoChevalier: Not related to RTL either (sorry should probably have done separate PRs)
This is to fix long words that are cut. Footer of https://donate.mozilla.org/de/one-page, should be “Datenschutzerklärung”:
capture d ecran 2017-01-24 a 01 44 31

ScottDowne: Cool, thanks for clearing that up.

less/shared.less
ScottDowne: That gradient shouldn't be there at all. It's only happening for chrome and safari on a mac, not seeing it in windows or linux.

ScottDowne: we can just add `outline: none;` on our inputs to override whatever is adding that.

ScottDowne:
Looks good to me.

PR #1732, opened 2016-12-31
Ensure Stripe isn't blocked.

ScottDowne:


src/components/payment-options.js
```diff
@@ -13,7 +13,7 @@ var PayPalButton = React.createClass({
     submit: React.PropTypes.array.isRequired,
     name: React.PropTypes.string.isRequired
   },
-  onChange: function() {
+  onClick: function() {
```
ScottDowne: This is for Paypal.

I switched Stripe button to click, and not change, because change wasn't working.

I figured keep Paypal in sync.

src/components/payment-options.js
```diff
@@ -88,7 +96,7 @@ var StripeButton = React.createClass({
     }
     return (
       <div className="half">
-        <input onChange={this.onChange} type="radio" className="payment-type payment-cc-input" name={name} value="cc" id={labelId}/>
+        <input ref={(input) => { this.input = input; }} type="radio" className="payment-type payment-cc-input" name={name} value="cc" id={labelId}/>
```
ScottDowne: This allows me to grab the element dom reference.

src/components/payment-options.js
```diff
@@ -67,18 +67,26 @@ var StripeButton = React.createClass({
...
+    this.props.onSubmit(this.props.validate, this.props.submit);
     reactGA.event({
       category: "User Flow",
       action: "Stripe Clicked"
     });
   },
+  componentDidMount: function() {
+    // Need this because Chrome on iOS plus StripeCheckout triggers a popup.
+    // We need to ensure the popup isn't blocked.
+    this.input.addEventListener("click", this.onClick);
```
ScottDowne: Need to do this directly on the DOM. I think React is doing some callbacks internally or something?

src/components/payment-options.js
```diff
@@ -67,18 +67,26 @@ var StripeButton = React.createClass({
...
+  },
+  componentWillUnmount: function() {
+    // Need to ensure we remove it, or else we'll have a memory leak.
+    // This is why you don't use DOM events in React.
+    this.input.removeEventListener("click", this.onClick);
```
ScottDowne: If react ever tears down this component, we need to ensure we also remove things we added.

src/components/payment-options.js
cadecairos: It think you're right

PR #1737, opened 2017-01-12
h1 fixes for jan-thank-you pages

beccaklam:
Fixed bug so that h1 displays the same on all three versions of the video landing page

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -10,7 +10,7 @@ var Index = React.createClass({
     var metaData = this.props.metaData;
     var robots = 'index, follow';
     var googleFonts = "https://fonts.googleapis.com/css?family=Open+Sans:600,400,300,300italic";
-    if (metaData.current_url.indexOf("jan-thank-you-cute") !== -1) {
+    if (metaData.current_url.indexOf("jan-thank-you") !== -1) {
       googleFonts = "https://fonts.googleapis.com/css?family=Roboto+Slab:600,400,300,200,100";
     }
 
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -10,7 +10,12 @@ var Index = React.createClass({
     var metaData = this.props.metaData;
     var robots = 'index, follow';
     var googleFonts = "https://fonts.googleapis.com/css?family=Open+Sans:600,400,300,300italic";
-    if (metaData.current_url.indexOf("jan-thank-you") !== -1) {
+    var tyPages = [
+      '/' + this.props.locale + '/jan-thank-you-cute',
+      '/' + this.props.locale + '/jan-thank-you-sincere',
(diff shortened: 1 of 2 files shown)


TheoChevalier:
This does not seem to work, it falls back to Georgia font

TheoChevalier:
Ok, I’ve found why it’s not working. @beccaklam you can do that in your patch https://github.com/mozilla/donate.mozilla.org/commit/9811c57aebddbdb639540a317629517ccbd986dd Kind of ugly, but it works for me locally

TheoChevalier:
Actually, we should find another way, http://localhost:3000/fr/jan-thank-you-superhero will work, this won’t http://localhost:3000/fr/jan-thank-you-superhero/ (trailing slash)

beccaklam:
@TheoChevalier It seems to work on my computer, but that's not a good indicator I'm guessing... I'm not exactly sure what we can do. Scott was the one who helped me write that bit as the JS is beyond what I can usually do...

cadecairos:
> Actually, we should find another way, http://localhost:3000/fr/jan-thank-you-superhero will work, this won’t http://localhost:3000/fr/jan-thank-you-superhero/ (trailing slash)

missed this! nice catch.

src/pages/index.js
```diff
@@ -10,7 +10,12 @@ var Index = React.createClass({
     var metaData = this.props.metaData;
     var robots = 'index, follow';
     var googleFonts = "https://fonts.googleapis.com/css?family=Open+Sans:600,400,300,300italic";
-    if (metaData.current_url.indexOf("jan-thank-you") !== -1) {
+    var tyPages = [
+      '/' + this.props.locale + '/jan-thank-you-cute',
+      '/' + this.props.locale + '/jan-thank-you-sincere',
+      '/' + this.props.locale + '/jan-thank-you-superhero'
+    ];
+    if (tyPages.indexOf(metaData.current_url) >= 0) {
```
cadecairos: you could try:
```js
if (metadata.current_url.match(/jan-thank-you-(cute|sincere|superhero)/) { /*...*/ }
```

cadecairos: which matches with/without trailing slash:

TheoChevalier:
My bad, this alone seem to work (seems I had cache issues):
`if (metaData.current_url.indexOf("jan-thank-you") !== -1) {`

beccaklam:
@cadecairos @TheoChevalier Ok I reverted the code in src/pages/index.js back to

>     if (metaData.current_url.indexOf("jan-thank-you") !== -1) {
>       googleFonts = "https://fonts.googleapis.com/css?family=Roboto+Slab:600,400,300,200,100";
>     }
> 

This is hopefully it!

PR #1739, opened 2017-01-21
Paypal locale codes are pretty busted - fixes #1738

### LINKED ISSUE #1738: Paypal locale codes are pretty busted
@TheoChevalier So I found something that's not good.

Paypal has what seems to be their own unique locale codes. See here: https://developer.paypal.com/docs/classic/api/locale_codes/

We initially handled this conversion here: https://github.com/mozilla/donate.mozilla.org/blob/master/intl-config.js#L4

However, as we added new locales, we never turned them on in paypal properly.

What this means is the login screen won't be in the right language for a lot of our locales. Once the user logs in, they get their user set language.

To test, try making a paypal donation in /de/ the paypal login screen should be in german. Then try in /pl/ and the login screen is in English, and should be in Polish.

TheoChevalier:


ScottDowne:
Do you suspect the Travis failure is related to your patch? I don't see how it could be related... I feel like Travis is borken.

TheoChevalier:
Looks like it was the usual SNAFU, restarted the job, seem to pass now

ScottDowne:
I noticed you're missing Polish. Might be missing others? But I stopped checking at that point.

You have a trailing comma on the last line of the JSON object.

These are the reasons for the R-

Some other potential updates:

Also, I wouldn't say no to a automated test that runs all the locales through the `paypalLocales` object to see if any are missing. That way if we add a new locale, it'll start to fail the unit tests until we add it to `paypalLocales`. But we can do that in a followup patch.

I also wouldn't say no to moving the `paypalLocales` out of the intl-config file and into the `src/data/locale-data` file, or otherwise find a way to merge that data. It's likely we miss keeping one of these data sources up to date if they are separate, but again, we can do that in another patch too.

ScottDowne:
@TheoChevalier ping, in case you missed my comment.

TheoChevalier:
Sorry, hadn’t had much time yesterday, looking

TheoChevalier:
You’ve made good points about tests and the location, and yeah, let’s do that in another PR so that we can fix the main issue faster

PR #1743, opened 2017-02-14
Arabic currencies

TheoChevalier:
Arabic translation has been completed https://pontoon.mozilla.org/ar/fundraising/ (woop, woop!), lets support currencies for Arabic-speaking countries.

+ Adding default currency for Kabyle. Not adding one for Arabic though, none really make sense as default


ScottDowne:
I think this is fine, from a technical standpoint. One small nit.

I am however worried about the UX concerns around the dropdown. It's just a simple dropdown, nothing fancy, and might make it too hard for the user to find their currency, for any currency. Making it pretty counter productive. Possibly.

Don't think that should stop us from landing, but it might become an issue very quickly.

TheoChevalier:
Right, that’s a real concern. I am wondering if inverting currency code and symbol would make the list easier to read. Using monospaced font would be even better, but I’m not sure if we can change it in a native select list, I guess it’s using OS styling, including the font?

src/data/currencies.js
```diff
@@ -293,6 +343,16 @@ module.exports = {
       monthly: ['40', '20', '10', '7']
     }
   },
+  'qad': {
```
TheoChevalier: good catch, should be qar, with a different symbol too

TheoChevalier:
Since https://donate.mozilla.org/ar is now live, we should probably merge & push this one too. @ScottDowne sounds good? (I’ve fixed the qad mishap)

ScottDowne:
Yeah, I think this is fine.

I didn't take the time to go through every currency to make sure the conversion rates make sense...

PR #1746, opened 2017-02-23
Add key prop to video thank you page.

ScottDowne:


ScottDowne:
@TheoChevalier This should fix the issue you found in https://github.com/mozilla/donate.mozilla.org/pull/1741

TheoChevalier:
Woops, looks good

PR #1767, opened 2017-07-28
Update currencies

TheoChevalier:
What this PR does:

- Reversing the order in the drop down for improved readability. Comparison:
capture d ecran 2017-07-26 a 16 18 02capture d ecran 2017-07-26 a 16 16 36

- Setting currencies as default  for more locales, looks like I previously missed some 👎
- Supporting a few more currencies. Pretty sure I checked for the Turkey currency a few month ago and it wasn’t supported, so it looks like Stripe added support for a bunch of new currencies :) In the case of `myr`, it is also supported by PayPal according to their doc, but I’m getting a `Undefined token` error. Not sure if it’s because something needs to be flipped on PayPal side or what not, so I’m not enabling it for PayPal yet.
- Adjusting presets for a few currencies for consistency


ScottDowne:
This all looks pretty logical to me. r+

TheoChevalier:
Thanks!

PR #1770, opened 2017-07-31
Updating some thank you strings.

ScottDowne:
Fixing https://github.com/MozillaFoundation/Advocacy/issues/1016

### DIFF
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -102,6 +102,6 @@ i_donated_to_thunderbird=I donated to @mozthunderbird today to #freetheinbox. Jo
 support_mozilla=Support Mozilla
 
 # Title for January Thank You pages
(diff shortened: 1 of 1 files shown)


ScottDowne:
@TheoChevalier @WillatMozFdn

TheoChevalier:
Aren’t we getting a whole new copy? In this case, it would make more sense to wait for the new copy instead of translating the strings twice. If @WillatMozFdn needs the strings right away, maybe we can just search and replace `2016` with `2017`, then localize the brand new copy?
How does that sound?

ScottDowne:
I thought just removing the "in 2016" bit was all we were doing. If there is new copy, then yes, I agree with you. Let me double check.

ScottDowne:
Double checked, we're good on this change for the next several months.

PR #1771, opened 2017-07-31
Expose the email subscription form to all relevant locales

TheoChevalier:
As discussed with @WillatMozFdn, let’s expose the subscription form to more locales than en-* and de.
We’re not including Italian because the MoCo list does not support it. I assume this is the only place that needs to be updated. I can’t test locally if I actually get subscribed, but this at least displays the form correctly.

### DIFF
--- a/src/pages/thank-you.js
+++ b/src/pages/thank-you.js
@@ -17,7 +17,7 @@ var ThankYou = React.createClass({
   render: function() {
     var className = "row thank-you-page";
     var signUpOrSocial = (<Social/>);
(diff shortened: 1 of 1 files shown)


ScottDowne:
Locally the signup is using some signup staging, so you won't go onto the actual list locally.

not sure how robust the staging is.

ScottDowne:
This should be fine though. R+

TheoChevalier:
Yeah, that was my guess. Alright, thank you!

PR #1774, opened 2017-08-01
Fix text-transform in Turkish (Bug 231162)

TheoChevalier:
More background if you’re not familiar with the changes the browser is making depending on the value of the lang attribute (I was not): https://bugzilla.mozilla.org/show_bug.cgi?id=231162

Basically, before:
capture d ecran 2017-08-01 a 13 01 04

After the fix, using a Turkish browser (note the last character):
capture d ecran 2017-08-01 a 13 00 58

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -44,7 +44,7 @@ var Index = React.createClass({
       dir = 'rtl';
     }
     return (
-      <html dir={dir}>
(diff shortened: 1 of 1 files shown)


TheoChevalier:
Turkish localizer confirmed this is fixing the issue, merging this easy one

PR #1775, opened 2017-08-02
Update about.js

vaggelisifa:
A slight change of `<span>` to `<p>`


TheoChevalier:
Thanks for your patch, unfortunately I don’t think we can merge it. Unless I’m missing something, we don’t want the text to go on a new line here. Let me know if you had something different in mind

PR #1781, opened 2017-09-02
Fix #1569 - Cancel Thunderbird donation in paypal sends user back to …

### LINKED ISSUE #1569: Cancel Thunderbird donation in paypal sends user back to Mozilla donation page instead of Thunderbird donation page
From https://donate.mozilla.org/en-US/thunderbird/about/ I start a donation for paypal and am sent to https://www.paypal.com/cgi-bin/webscr?cmd=_express-checkout&useraction=commit&token=EC-2LC32813TA104134A#/checkout/login

If I then click "Cancel and return to Mozilla Foundation" which is https://donate.mozilla.org/?token=EC-2LC32813TA104134A ... I then get sent to https://donate.mozilla.org/en-US/

Desired results - I should be returned to https://donate.mozilla.org/en-US/thunderbird/about/

mkmelin:
…Mozilla donation page instead of Thunderbird donation page.

I think this should fix it, but it seems testing checkout on localhost doesn't work  so I can't really verify it. Maybe there's some env variable that I would have to set?

### DIFF
--- a/routes/paypal.js
+++ b/routes/paypal.js
@@ -18,6 +18,10 @@ function setupPaypal(transaction, callback) {
 
   var returnUrl = transaction.serverUri + '/api/paypal-redirect/' + frequency + '/' + locale + '/' + appName + '/' + accountType + '/';
 
+  var cancelUrl = transaction.serverUri + '/';
+  if (appName === 'thunderbird') {
+    cancelUrl += 'thunderbird/';
+  }
   var charge = {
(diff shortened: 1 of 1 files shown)


TheoChevalier:
It looks good here, thanks!
Regarding env config, if you’ve copied https://github.com/mozilla/donate.mozilla.org/blob/master/sample.env into `.env`, it should work locally

mkmelin:
Yeah, that was what was needed. It's working locally now.

PR #1783, opened 2017-09-03
adding thumbnail image with License to README.mm

tmm88:
**this PR does basically aim at:**
- *adding a thumbnail image with license to README.md*

---

**motivation for that does basically with making file**
- *more more informative, simpler, easier, and more direct to be read*


TheoChevalier:
Hi, thanks for your contribution. However, I don’t think we can merge this Pull Request for the following reasons:

 - We generally don’t want to link to unknown, external domains
 - The branding in the picture is outdated
 - Adding such pictures for License information is not a common practice at Mozilla
 - The license info is already prominent (and accessible to screen readers) at the top of the page

Thanks!

PR #1786, opened 2017-09-18
Mofo faq update

TheoChevalier:
I’m temporarily duplicating the template in both faq.js files, in order to expose new content in English right away while the strings are being translated.

The second commit updates the Privacy Policy link for all locales, trying to reduce the burden for localizers. Especially since Pontoon’s translation memory won’t kick in for strings over 255 bytes.


ScottDowne:
I just saw this now. Is this something you want me to start reviewing?

TheoChevalier:
@ScottDowne There still a few string changes pending (Audrey has updates, waiting copy for the new SEPA answer, etc), but most of it is ready for review I think, yeah. Thanks :)

ScottDowne:
OK great, after taking a quick look, it actually looks pretty straightforward to me. We can wait for the final strings though if we want.

TheoChevalier:
Looks like React is yelling in the console.

`Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:`
 `(client)  data-reactid="24"><b data-reactid="25">`
 `(server)  data-reactid="24"><span data-reactid="2`

It doesn’t like `<b>` tags wrapping FormattedHTMLMessage:
`<b><FormattedHTMLMessage id='receiving_bank' /></b>`

Not sure how to fix this… Tried with `<span>` + CSS class, same warning.

@ScottDowne Do we need to fix this?

ScottDowne:
It's usually fine. I can take a look in a bit to be sure.

ScottDowne:
I don't see the checksum invalid warning. My guess is you need to do a hard refresh, and if that doesn't work, try rebuilding the server with npm install.

When you do a server side build with react, the client still checks to see if anything is different in the client to what it expects from the server, if there are no changes, react then does nothing. If anything isn't the same, react updates it and gives this warning. It's harmless.

My guess is the browser has cached an older page, and is actually rendering that instead of what the server sent and what react expects. And react then needs to go in and update, thus undoing the browser cache.

Let me know if you think that adds up.

TheoChevalier:
Aaah, thanks @ScottDowne ! That was it, I thought using `npm start` was hard enough.

Also, https://xkcd.com/1854/.

I will fix the conflict I just created a bit later then merge and send communication to localizers, I think it’s ready!

PR #1789, opened 2017-09-21
Change how we capture and report monthly Stripe charges

cadecairos:
Previously, we only captured the first charge related to a subscription, and sent data about the subscription to basket. 

This PR moves that code to the webhook we have set up for updating charges related to subscriptions, letting us not only capture the initial subscription charge, but any and all subsequent charges. I also now pass the charge id as `transaction_id` and pass the subscription id as `subscription_id` so that it can be properly filtered in saleforce when we're doing reporting.

cc/ @pmac @stephaniemcv

### DIFF
--- a/routes/index.js
+++ b/routes/index.js
@@ -250,20 +250,6 @@ var routes = {
                 customer_id: customer.id
               });
 
-              basket.queue({
-                event_type: "donation",
-                last_name: customer.sources.data[0].name,
-                email: customer.email,
-                donation_amount: basket.zeroDecimalCurrencyFix(subscription.quantity, subscription.plan.currency),
-                currency: subscription.plan.currency,
-                created: subscription.created,
-                recurring: true,
-                frequency: "monthly",
-                service: "stripe",
-                transaction_id: subscription.id,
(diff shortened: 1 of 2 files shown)


routes/index.js
```diff
@@ -609,6 +598,22 @@ var routes = {
...
+              email: subscription.customer.email,
+              donation_amount: basket.zeroDecimalCurrencyFix(charge.amount, charge.currency),
+              currency: charge.currency,
+              created: charge.created,
+              recurring: true,
+              frequency: "monthly",
+              service: "stripe",
+              transaction_id: charge.id,
+              subscription_id: subscription.id,
+              project: charge.metadata.thunderbird ? "thunderbird" : ( charge.metadata.glassroomnyc ? "glassroomnyc" : "mozillafoundation" )
```
cadecairos: @ScottDowne is this going to work still? I can't remember if the metadata is applied to the subscription or the charge..

cadecairos: (It was wrong, I updated it to check `updateData`)

cadecairos:
@ScottDowne 🏓

ScottDowne:
R+

PR #1809, opened 2017-10-16
Adding signatures to stripe webhooks

### LINKED ISSUE #1766: Verify Stripe Webhook Signatures
We use tokens to ensure we don't receive ust any old POST request, but we should also verify the signature that Stripe sends with each hook: https://stripe.com/docs/webhooks#signatures

patjouk:
Fixes #1766 

I also added two missing semicolons to make the linter happy.

Todos:
- [x] Move signature code to stripe.js
- [x] Bump stripe module version to v5: I need to upgrade to at least 4.19 to use the webhooks signature feature but let's try moving to v5 and see if break stuff or not.
- [x] Tests

Deploy todos:
- [x] Clean up the test webhooks I've setup
- [x] Setup signature validation for `stripe-charge-failed` and `stripe-charge-refunded`
- [x] Add the new env variables for both staging and deploy stacks (`STRIPE_WEBHOOK_SIGNATURE_CHARGE_SUCCESS` and `STRIPE_WEBHOOK_SIGNATURE_CHARGE_FAILED`, `STRIPE_WEBHOOK_SIGNATURE_CHARGE_REFUNDED`, `STRIPE_WEBHOOK_SIGNATURE_DISPUTE`). ⚠️ Each webhook has its own signature key ⚠️

### DIFF
--- a/routes/index.js
+++ b/routes/index.js
@@ -463,7 +463,15 @@ var routes = {
     }
   },
   'stripe-charge-failed': function(request, reply) {
-    var event = request.payload;
+    var endpointSecret = process.env.STRIPE_WEBHOOK_SIGNATURE_CHARGE_FAILED;
+    var signature = request.headers["stripe-signature"];
+
+    var event;
+    try {
+      event = stripe.constructEvent(request.payload, signature, endpointSecret);
+    } catch (constructEventErr) {
+      return reply(boom.forbidden('An error occurred while verifying the webhook signing secret', constructEventErr));
+    }
 
     if (event.type !== 'charge.failed') {
       return reply('This hook only processes charge failed events');
@@ -480,7 +488,15 @@ var routes = {
     return reply("charge failed event processed");
   },
   'stripe-charge-refunded': function(request, reply) {
-    var event = request.payload;
+    var endpointSecret = process.env.STRIPE_WEBHOOK_SIGNATURE_CHARGE_REFUNDED;
+    var signature = request.headers["stripe-signature"];
+
+    var event;
+    try {
+      event = stripe.constructEvent(request.payload, signature, endpointSecret);
+    } catch (constructEventErr) {
+      return reply(boom.forbidden('An error occurred while verifying the webhook signing secret', constructEventErr));
+    }
 
     if (event.type !== 'charge.refunded') {
       return reply('This hook only processes charge.refunded events');
--- a/routes/index.js
+++ b/routes/index.js
@@ -590,11 +590,10 @@ var routes = {
     var endpointSecret = process.env.STRIPE_WEBHOOK_SIGNATURE_CHARGE_SUCCESS;
     var signature = request.headers["stripe-signature"];
 
-    var event;
-    try {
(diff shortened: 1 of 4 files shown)


routes/index.js
```diff
@@ -369,7 +369,7 @@ var routes = {
 
           request.log(['paypal', 'checkout', frequency], log_details);
 
-          var timestamp = new Date(data.txn.PAYMENTINFO_0_ORDERTIME).getTime() / 1000
+          var timestamp = new Date(data.txn.PAYMENTINFO_0_ORDERTIME).getTime() / 1000;
```
cadecairos: 👍 💯

patjouk:
I've done a few tests (manual and automatic) after moving to stripe v5: everything seems to be running okay.

This PR is now ready for review @cadecairos 

Deploy todos:
- [ ] Clean up the test webhooks I've setup
- [ ] Add the new env variables for both staging and deploy stacks (`STRIPE_WEBHOOK_SIGNATURE_CHARGE_SUCCESS` and `STRIPE_WEBHOOK_SIGNATURE_DISPUTE`). ⚠️ Each webhook has its own signature key ⚠️

routes/stripe.js
```diff
@@ -89,6 +89,9 @@ var stripeRoutes = {
   },
   retrieveSubscription: function(customerId, subscriptionId, options, callback) {
     stripe.customers.retrieveSubscription(customerId, subscriptionId, options, callback);
+  },
+  constructEvent: function(payload, signature, endpointSecret) {
```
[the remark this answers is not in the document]
cadecairos: actually, nevermind about this request, I realized just now that the event object we return is used in the request, so this does make sense.

patjouk:
I tried to keep the error message using a list but I'm sure there's a more elegant way to do it :D

routes/stripe.js
```diff
@@ -89,6 +89,17 @@ var stripeRoutes = {
...
+    var event;
+
+    try {
+      event = stripe.webhooks.constructEvent(payload, signature, endpointSecret);
+    } catch (constructEventErr) {
+      event = [null, constructEventErr];
```
cadecairos: we could do 

```js
server.log('error', constructEventErr);
return null;
```

Then in the webhook handlers:

```js
event = stripe.constructEvent(request.payload, signature, endpointSecret);

if (!event) {
  return reply(boom.forbidden('An error occurred while verifying the webhook signing secret'));
}
```

routes/stripe.js
```diff
@@ -96,7 +96,7 @@ var stripeRoutes = {
     try {
       event = stripe.webhooks.constructEvent(payload, signature, endpointSecret);
     } catch (constructEventErr) {
-      event = [null, constructEventErr];
+      console.error('Error: ', constructEventErr);
```
cadecairos: Can we make it `"constructEvent Error:"`, that way it's more easily surfaced in the logs when searching

patjouk:
I've added the signature keys. Ready to merge @cadecairos ?

PR #1829, opened 2017-10-24
Adding Redux Persist.

ScottDowne:


ScottDowne:
@cadecairos noticed a weird bug, which I swore I fixed, but I guess I didn't. This patch fixes it.

Anyway, if you don't do a hard refresh some state can get out of sync with browser cached input values.

This isn't a huge issue in most cases, but where it is likely the most problematic and why I fixed it asap is this:

1. Load current donate form from donate.mozilla.org
2. select monthly
3. refresh
4. after the refresh select an amount
5. hit donate with stripe

Expected: should have a monthly donate popup
Actual: we've downgraded to one time.

This hasn't been around for too long, it started some time after I added redux, but isn't actually related to redux, but a browser update. I think.

Either way, the way this fixes it is forces the state to be in the same as the previous load, which mimics what cache is doing, which essentially puts them in sync. Thoughts?

ScottDowne:
Chris looked at the recent change to switching to session storage, and was fine with it too. Merging.

PR #1833, opened 2017-10-31
[WIP] refactored the server code so that it's easier to understand

Pomax:
this PR puts the server code in its own directory, and splits up server.js into more manageable modules. It also updates `single-form.js` so that the render function is easier to understand.

tests applied to verify this all went fine:
- ran `npm install`, no errors
- ran `npm test`, no errors
- the site looks fine, but this needs an extra pair of eyes before I will trust it's a safe refactor.

please verify that the changes from https://github.com/mozilla/donate.mozilla.org/commit/794e28ff7f4bcdb5c17aa522939b155f0e3f6ce5 are preserved in this refactor (I had to rebase and while I'm pretty sure I made all the changes from that commit made it into this one, that should be verified!)


src/components/single-form.js
```diff
@@ -71,65 +71,6 @@ var singleForm = React.createClass({
     }
     return true;
   },
-  renderPaymentOptions: function() {
```
ScottDowne: Your call, but if it's possible to move these changes into another PR, I could probably get to reviewing just the changes to single-form.js much much faster.

Pomax:
proooobably superceded by https://github.com/mozilla/donate.mozilla.org/pull/1834 but keeping these files around as this was the refactor prior to stripe changes rebase

src/components/single-form.js
Pomax: can do.

Pomax:
superceded

Issue #1811, opened 2017-10-16
Setup suggested recurring donation amounts based on one-time donation


alanmoo:
This may be in the original epic ticket, but we should confirm them here

WillatMozFdn:
yes, copied over from the original epic:
General parameters:

    the one-time donation is processed first, so we know we have that transaction processed already

    if someone donates 0-14.99$, we do NOT fire the ask for a monthly.

    if someone donates $15 to $34.99, we suggest $5 monthly

    if someone donates $35-59.99, we suggest $10 monthly

    $60-119.99, suggest $15 monthly

    $120-$250, suggest $20 monthly

    above $250 suggest $30 monthly

Ideally, we would set the monthly to start not immediately, but one month from today. (If that's not possible, we just start the monthly right away.)

alanmoo:
We should also discuss how these amounts convert to reasonable suggestions with different currencies.

WillatMozFdn:
Hey Alan,

Would it help if we restricted this down to 2-3 locales/currencies?  I'm
thinking EN-US ($), DE (euros) and FR (also euros).  Based on some recent
snippet tests, I think this will get us a very high percentage of what we
need ... and hopefully simplify this question, since the amounts for
dollars and euros will be very similar if not the same.

Let me know what you think, thanks!

 - Will

alanmoo:
Sounds reasonable to me. Cc @pomax

gideonthomas:
@WillatMozFdn, to clarify, should we go ahead and just use the exact same amounts for DE (€) and FR (€) as EN-US ($)?

That is to say for EN-US, FR, and DE:

| One-time Donation ($ or €) | Suggested Monthly Donation ($ or €) |
|---|---|
| 0 - 14.99 | - |
| 15 - 34.99 | 5 |
| 35 - 59.99 | 10 |
| 60 - 119.99 | 15 |
| 120 - 250 | 20 |
| above 250 | 30 |

WillatMozFdn:
@gideonthomas ... yes.  Let's just use the exact same amounts in Euros and dollars.

[...]

PR #1835, opened 2017-10-31
Monthly prompt

Pomax:
This updates the `thank you` page such that when a one time donation comes in, we show a text elliciting users to perform a monthly donation as well. (test by hitting up https://donate-mozilla-org-us--pr-1835.herokuapp.com/en-us/thank-you/?payment=Stripe&str_amount=2000&str_currency=usd&str_id=hk_lollercake&str_frequency=one-time&email=theboss@mofo.ca&country=Canada for instance)

The donation page, when opened through that link, should show a text that can be tailored for users that specifically opted into a monthly donation on top of the donation they just made. (test this with https://donate-mozilla-org-us--pr-1835.herokuapp.com/?fd=true&frequency=monthly&currency=usd&amount=2.5, the `fd=true` triggers the custom message).

Note that we should *probably* not merge this into master until we've figured out the exact:
- message on the thank you page
- message on the donate page
- styling on both pages
- exact conversion function(s) for turning one-time donated amount into recurring monthly charges


src/pages/thank-you.js
```diff
@@ -4,27 +4,68 @@ import Signup from '../components/signup.js';
...
+*/
 
 var ThankYou = React.createClass({
   contextTypes: {
     intl: React.PropTypes.object
   },
+  componentWillMount: function() {
+    this.monthlyInvitation = null;
+
+    let query = locationSearchParser(window.location);
```
ScottDowne: Doesn't this break on the server side render?

src/pages/thank-you.js
```diff
@@ -4,27 +4,68 @@ import Signup from '../components/signup.js';
...
+    if (query && query.str_frequency === "one-time") {
+      let currencyCode = query.str_currency;
+      let trueAmount = amountModifier.reverse(
+        query.str_amount,
+        query.payment.toLowerCase(),
+        currencyCode
+      );
+      let suggestion = suggestMonthly(trueAmount);
+      console.log(suggestion, trueAmount);
```
ScottDowne: Just a friendly reminder to remove this. Otherwise carry one :)

src/pages/thank-you.js
Pomax: it does not - I expect because of the polyfills.

src/pages/thank-you.js
Pomax: yup, WIP - all console.logs will get cleaned up!

alanmoo:
@WillatMozFdn - this is still in progress, but the main workflow is there. Can you check out the link in the first post and make sure that's what you were expecting from a functionality standpoint (dollar amounts still need to be worked in) before we get some designer guidance?

gideonthomas:
Let's chat quickly about this during the end of our planning meeting but just some food for thought before that for @ScottDowne and @Pomax

As per the original ticket, the suggestion was to make the UX similar to https://cstage-zz.trilogyinteractive.com:10444/test-upsell.html?show_amount=true. I think what's been done so far in this PR is great!

One big distinction between the demo linked and the functionality that this PR introduces is the number of interactions the user has with the UI to finish the monthly donation.

**Current Behavior**
1. Finish paying one time donation (with payment processing finished on the backend)
2. Redirect to thank you page with "Your one-time payment was successful" and two CTAs - Subscribe to mailing list and Make Monthly Donation
3. Click on Make Monthly Donation and then see the donation form again with pre-filled amount
4. Click on Pay with Credit Card and see information form
5. Enter all the information again and click "Pay"
6. Redirect to thank you page with one CTA - Subscribe to mailing list

**Alternative Behavior**
1. Finish paying one time donation (with payment processing finished on the backend)
2. Redirect to thank you page with "Your one-time payment was successful" and two CTAs (design can suggest how we deal with this) - Subscribe to mailing list and Make Monthly Donation **of $xx.xx**
3. Click on Make Monthly Donation
   - This would trigger a call to the backend to process the monthly subscription since we already know all of the information
   - My thought behind this is that in the reply we get back from "finishing one-time payment", found [here](https://github.com/mozilla/donate.mozilla.org/blob/master/server/routes/index.js#L187-L195) seems to include all of the information we need to hit up the "recurrent payment" API endpoint again. So if we can use that, it doesn't seem like we need any input from the user other than the explicit "Make Monthly Donation" click.
4. Redirect to (or re-render the) thank you page with "You have been subscribed to monthly payments" and one CTA - subscribe to mailing list

I think before we try jumping to the alternative behavior, my question is how much work (how many lines of code / how complicated / how many days of work, whatever you prefer) do you think it will be to make this change? Please also feel free to point out any incorrect assumptions I've made or to suggest something different.

Note that this will also inform the design follow-up that needs to be done.

WillatMozFdn:
@gideonthomas -- Alan and I will be talking to the Trilogy guys on Monday to see if we can't set up a system/interface closer to theirs.  If possible, I would like this to flow like your "suggested behavior" steps, except that for steps #2 and 3 we stay in the modal window to make the ask for the monthly contribution (while the one-timer is processed in the background).  Then, after either a monthly is processed OR they decline the monthly, we kick them out to the thank-you page, where the primary CTA is to sign up for our mailing list.  Hope that makes sense?  Alan or I can update on Monday as to whether this is feasible under our current setup.  Thanks!

src/lib/suggest-monthly.js
```diff
@@ -0,0 +1,7 @@
+module.exports = function suggestMonthly(amount) {
+  // TODO: refine this function to actually do what
+  // we want it to do based on the amount donated,
+  // and the numbers we came up with.
+
+  return (1.5 * parseInt(amount, 10)/12) + "";
```
Pomax: update this based on https://github.com/mozilla/donate.mozilla.org/issues/1811

gideonthomas:
@ScottDowne, if you're continuing this patch, one thing to note is that when testing this, I noticed that when I was redirected back to the donation page to make a monthly donation, the amount was not pre-filled and the "Monthly" radio button was not selected.

gideonthomas:
@Pomax what would you like to do with this PR? Should it be closed or are there fixes in this that need to happen anyway (and we should extract them out of this PR and open a new one)?

ScottDowne:
I've extracted some of the stuff I directly needed into my PR.

I dunno what's left, but we can probably close this.

We can always find this PR if we close it if we decide we want more from it.

gideonthomas:
Yeah, but at some point we should delete this branch since it's on mozilla/donate.mozilla.org

PR #1844, opened 2017-11-02
Make review apps work. Maybe.

### LINKED ISSUE #1842: Enable review apps
It would be good (especially for designers) to test PRs without having to setup donate locally. Let's enable review apps for this repo.

This is definitely a nice to have and a lower priority than implementation issues.

alanmoo:
I think this closes #1842

### DIFF
--- a/app.json
+++ b/app.json
@@ -114,9 +114,6 @@
   },
   "formation": {
   },
-  "addons": [
-    "adept-scale"
-  ],
   "buildpacks": [
     {
       "url": "heroku/nodejs"
--- a/app.json
+++ b/app.json
@@ -4,10 +4,10 @@
   },
   "env": {
     "ADEPT_SCALE_LICENSE_KEY": {
-      "required": true
+      "required": false
     },
     "ADEPT_SCALE_URL": {
-      "required": true
+      "required": false
     },
     "APPLICATION_URI": {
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -4,10 +4,10 @@
   },
   "env": {
     "ADEPT_SCALE_LICENSE_KEY": {
-      "required": true
+      "required": false
```
gideonthomas: you should just remove these env vars because they aren't actually a part of staging.

alanmoo:
Ok, removed them.

app.json
```diff
@@ -3,12 +3,6 @@
   "scripts": {
   },
   "env": {
-    "ADEPT_SCALE_LICENSE_KEY": {
```
gideonthomas: also remove the adept scale add on from this json file as well

app.json
```diff
@@ -3,12 +3,6 @@
...
-      "required": true
-    },
-    "ADEPT_SCALE_URL": {
-      "required": true
-    },
     "APPLICATION_URI": {
```
gideonthomas: this probably needs to be the review app application uri (i don't know if we rely on this for cors, etc.) but I don't know how to set that here. I think for now maybe we can just leave this as you have it.

PR #1857, opened 2017-11-08
First pass implementation of a SEPA donation processing endpoint

cadecairos:
Notes:

- Not tested
- We need more information about when to send donation receipts (time of charge creation, or when funds successfully transfer)
- Did not implement monthly logic.


cadecairos:
My reaction when I realize we run donate on node 4.8.x:

tableflip

PR #1858, opened 2017-11-08
[WIP] add SEPA payment as a third donation option

Pomax:


package.json
```diff
@@ -4,11 +4,13 @@
...
   "main": "start.js",
   "scripts": {
-    "start": "npm-run-all --parallel build:l10njson watch:js server",
+    "start": "npm run clean && npm-run-all --parallel build:l10njson watch:js server",
+    "clean": "npm run rebuild && shx rm public/main*.js && shx rm public/style*.css",
     "build:l10njson": "pontoon-to-json --dest ./public",
     "build:exchangerates": "node scripts/exchangerates.js",
     "build:babel": "babel src --compact=true -s --out-dir dist",
     "build:webpack": "webpack --optimize-minimize --optimize-dedupe",
+    "rebuild": "rimraf ./dist && npm-run-all build:*",
```
gideonthomas: any reason we're using `rimraf` instead of `shx rm -rf`? `shx`/`shelljs` use `rimraf` internally as far as I can tell

Pomax: `shx rm -f` did not want to actually delete a directory on windows for me.

gideonthomas: what about with the `-r` flag as well?

src/pages/sepa.js
```diff
@@ -0,0 +1,145 @@
...
+            iban: this.state.iban
+          },
+          currency: 'eur',
+          owner: {
+            name: this.state.name
+          }
+        }).then( result => {
+          console.log(result);
+
+          if (result.error) {
```
gideonthomas: this should be an object in the form of https://stripe.com/docs/api/node#errors
where you would use the codes in https://stripe.com/docs/sources/sepa-debit#error-codes to determine what kind of error it was. 

You might want to check in with @ScottDowne about how we handle errors for card payments when creating a source object on the front-end (do we show a generic error message, etc.)

gideonthomas:
@Pomax, does this branch now work in such a way that hitting (either directly or by clicking on the SEPA button on the main donation page) `/sepa` is a network request to the server?

Pomax:
@gideonthomas it was already doing that: as switch-over it is a `window.location` change (so a real URL request, not an app-internal change with history fakery).

package.json
Pomax: haven't tried, but if shx already uses rimraf, then npm's deduplication should make using rimraf explicitly irrelevant.

gideonthomas:
@Pomax, if we merge @cadecairos patch first, he introduces the `ENABLE_SEPA` env var. Out here, if you bake that into webpack as well, we can share that env var feature flag between client and server.

Pomax:
@gideonthomas added. webpack, as well as several JSX components, use the `SEPA_ENABLED` feature flag now.

PR #1859, opened 2017-11-08
Update to node 8.9 LTS

cadecairos:


cadecairos:
Lets see if we can get a review app going

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- 4.8.4
+- 8.9.0
```
gideonthomas: why not just `- 8` since we're doing effectively that in the `package.json` file?

PR #1865, opened 2017-11-12
Sync the Thunderbird about page with general Mozilla changes

mkmelin:
I wanted to add the SEPA and wire transfer infos into the Thunderbird front page. 
This was easier after syncing with the changes that have been done to the general pages.


ScottDowne:
Looks good to me.

PR #1867, opened 2017-11-13
WIP: Style SEPA page

alanmoo:
The buttons need fixing (and only to show the SEPA button when Euros is selected)

We also need to determine what other text needs to be on `/sepa`


alanmoo:
@flukeout Can you take a look at the review app for styling guidance? There should be a "view deployment" button in the status update above this after a minute or two.

flukeout:
I'll file some QA issues related to this PR.

PR #1881, opened 2017-11-14
Payment type button updates

flukeout:
**Changes**
* Added a SEPA logo
* Fixed logo positioning on buttons
* Split the payment type logos into separate images instead of relying on the sprite sheet
  * This allowed me to remove a lot of CSS that dealt with logo positioning etc
* Rejigged the submitting state so that it fits the new three-across button layout
* Made the gear an SVG instead of a font-awesome icon so that I position it more easily so the spinning doesn't look wobbly

**Should appear like this...**

**Submitting state like this...**

cc @alanmoo


less/components/payment-buttons.less
```diff
@@ -49,100 +42,27 @@ input.payment-type:checked + label {
   background-image: linear-gradient(transparent, rgba(0, 0, 0, 0.2));
 }
 
-label.payment-paypal-label,
```
alanmoo: This deletion feels so good

flukeout: Haha, right? Good riddance spritesheet!

PR #1882, opened 2017-11-14
Hide SEPA button except for Euros

alanmoo:
@ScottDowne I didn't handle showing/hiding payment methods the same way you do elsewhere, but I think that's ok since this would be more of an "include" key I'd need to add than using the existing "exclude" key in `curriencies.js`. Thoughts?


flukeout:
Will review first thing tomorrow @alanmoo

flukeout:
@alanmoo Looks good, the SEPA options shows up when selecting Euros, as expected.
I found a problem related to some changes I made in a pervious PR however...

If you select the MXN$ currency, you will see the Credit Card button is missing a logo. It's trying to find..

* ``assets/images/payment-type-cards-no-amex.png``

However, the image file we have in our assets folder is called

* ``assets/images/payment-logos-cards-no-amex.png``

To fix this, just rename the file to the one it's looking for and we should be good.

flukeout:
Ready to merge @alanmoo - and let's file a follow-up on the missing CC icons that you pointed out.

PR #1883, opened 2017-11-15
Updates to input fields on SEPA form

### LINKED ISSUE #1873: Donate form QA: Un-needed padding on SEPA input fields.
If you select the SEPA option and click on one of the two input fields, you'll notice that there is some extra left padding...

image

Let's get rid of it.

### LINKED ISSUE #1874: Donate form QA: Focus the Name field on the SEPA form by default.
Based on [this deploy](https://donate-mozilla-org-us--pr-1867.herokuapp.com/)

When a user selects the SEPA payment option, and the form changes to reflect that, we should focus the Name field so that they can start filling it out right away.

The field in question...
image

### LINKED ISSUE #1876: Donate form QA: Update field labels and add placeholders to SEPA form
We may want to update the labels, and maybe add placeholders or helper text for these fields...

image

* **Name** should probably be changed to something like **Account Holder's Full Name**
  * @gideonthomas is there a spec for this somewhere?
* Do we want to add some helper text for the IBAN number? Like, what it stands for or how to figure out which one to use? 
  * Thinking of things like the popups / extra info that help you find your 3-digit security number on credit cards

### LINKED ISSUE #1879: QA: The SEPA donation form needs an email field
According to the SEPA Stripe docs, when creating a customer, we need to provide an email (so that we can send them their receipt and notifications about pre-authorized payments).

We need an email field in the SEPA form as well.

cc @flukeout

flukeout:
**Changes**
* Removed the unneeded left padding on the input fields
  * Fixes #1873
* Added an email field
  * Just chucked it into the form, not really validating it or anything like that for now
  * Fixes #1879
  * @gideonthomas let me know what you want me to do with this
* Updated field labels
  * Fixes #1876
* Focuses the first input field when the page loads
  * Fixes #1874
  * Kind of hacky because we don't get access to the `ref` of that field on `componentDidMount` so had to move it to `componentDidUpdate` with a flag

**Should appear like this...**

cc @gideonthomas

### DIFF
--- a/less/shared.less
+++ b/less/shared.less
@@ -275,7 +275,7 @@ input[type="email"] {
   margin-bottom: 12px;
   border: 3px solid #EAEFF2;
   border-radius: 3px;
-  padding: 6px 7px 6px 33px;
+  padding: 6px 7px 6px 7px;
   background-color: #F7F9FA;
   outline: none;
 }
--- a/src/pages/sepa.js
+++ b/src/pages/sepa.js
@@ -34,8 +34,8 @@ var SEPA = React.createClass({
     script.onload = (e) => this.setState({ stripeLoaded: true });
     head.appendChild(script);
   },
-  componentDidUpdate: function(){
-    if(!this.state.initialFieldFocused) {
+  componentDidUpdate: function() {
+    if (!this.state.initialFieldFocused) {
       this.setState({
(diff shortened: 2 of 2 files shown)


gideonthomas:
@flukeout feel free to merge this in when you're ready

PR #1900, opened 2017-11-15
Email field fix

### LINKED ISSUE #1899: QA: Email field on thank-you page needs left-padding.
When I removed the global left padding on input fields, I inadvertently affected this page as well, causing the bug below, on the `thank-you` page...

image

flukeout:
On the `thank-you` page, the email field has the correct left padding now, which leaves room for the envelope icon, like this...

Fixes #1899


alanmoo:
Should this be against the SEPA branch?

flukeout:
Oh yeah something went wrong here for sure @alanmoo

PR #1909, opened 2017-11-16
Capture and record recurring txn failures properly

### LINKED ISSUE #1880: SFDC Upsert failing for failed recurring transactions
For recurring donations that fail, the upsert is failing in SFDC b/c it's missing the Close Date and StageName values. What is happening is that the transaction ID is not found in SFDC, so a new record is attempting to be created, but the fields needing to create a new record are not included in the data coming in from Stripe.

(1) We need to make sure that only specific failures are coming through Stripe 
(2) The same fields used to create a new donation Oppty record in SFDC should be passed over for these failures

cadecairos:
Fixes #1880

### DIFF
--- a/server/routes/webhooks/stripe-charge-failed.js
+++ b/server/routes/webhooks/stripe-charge-failed.js
@@ -35,7 +35,7 @@ let stripeChargeFailed = (request, reply) => {
     return reply(incorrectEventError);
   }
 
-  let {charge} = event.data.object;
+  let charge = event.data.object;
 
   // If the charge has no invoice ID, this is not a recurring charge
   if (!charge.invoice) {
--- a/server/routes/webhooks/stripe-charge-failed.js
+++ b/server/routes/webhooks/stripe-charge-failed.js
@@ -2,8 +2,14 @@ const boom = require('boom');
 const stripe = require('../stripe');
 const basket = require('../../lib/basket-queue');
 
+// values that never change between webhook requests
 const signature_header = 'stripe-signature';
 const event_type = 'charge.failed';
+const frequency = 'monthly';
+const recurring = true;
+const service = 'stripe';
+
+// Error/Success message constants
 const forbidden_error = 'An error occurred while verifying the webhook signing secret';
 const incorrect_event_error = 'This hook only processes charge failed events';
 const expand_charge_error = 'An error occurred while expanding the charge';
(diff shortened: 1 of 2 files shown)


server/routes/webhooks/stripe-charge-failed.js
```diff
@@ -0,0 +1,106 @@
...
+  if (!event) {
+    return reply(boom.forbidden(forbiddenError));
+  }
+
+  // Only process charge.failed webhook events
+  if (event.type !== event_type) {
+    return reply(incorrectEventError);
+  }
+
+  let {charge} = event.data.object;
```
gideonthomas: why do we destructure `event.data.object`? Is there a `charge` property on `event.data.object`?

server/routes/webhooks/stripe-charge-failed.js
```diff
@@ -0,0 +1,106 @@
...
+
+  // If the charge has no invoice ID, this is not a recurring charge
+  if (!charge.invoice) {
+    return reply(notRecurringCharge);
+  }
+
+  // Expand the invoice object on the charge
+  stripe.retrieveCharge(charge.id, (fetchChargeError, charge) => {
+    if (fetchChargeError) {
+      return reply(boom.badImplementation(`${expandChargeError}: ${fetchChargeError}`));
```
gideonthomas: one thing to note here is that if `fetchChargeError` is an object and you use string interpolation, it will show up as `[object Object]`

server/routes/webhooks/stripe-charge-failed.js
```diff
@@ -0,0 +1,106 @@
...
+    if (!charge.invoice || !charge.invoice.subscription) {
+      return reply(notPartOfSubscription);
+    }
+
+    let {customer, subscription} = charge.invoice;
+
+    // Expand the subscription object
+    stripe.retrieveSubscription(customer, subscription, (retrieveSubscriptionError, subscription) => {
+      if (retrieveSubscriptionError) {
+        return reply(boom.badImplementation(`${retrieveSubscriptionErrorMsg}: ${retrieveSubscriptionError}`));
```
gideonthomas: same note here

server/routes/webhooks/stripe-charge-failed.js
```diff
@@ -0,0 +1,106 @@
...
+      }
+
+      // Capture the variables we need to pass to the CRM
+      // Since this charge failed, we need to pass enough information
+      // for the CRM to create a new Opportunity record, and mark it
+      // immediately as 'Lost'
+      let {
+        currency,
+        created,
+        id: transaction_id,
```
gideonthomas: why destructure into `id` instead of leaving it as `transaction_id`? It doesn't seem like you use `id` anywhere but you redeclare it below and reassign it to `subscription_id`

server/routes/webhooks/stripe-charge-failed.js
cadecairos: Good catch

server/routes/webhooks/stripe-charge-failed.js
cadecairos: I can confirm this in the Stripe code, but I'm almost certain this will be an `Error`, and will have a useful toString implemented for outputting

server/routes/webhooks/stripe-charge-failed.js
cadecairos: You've got your understanding of how this destructuring assignment works:

This line of code will look for `id` on the right operand (`charge`), and assign it's value to `transaction_id`. This is understandably confusing, especially if you try to liken this syntax to that of Object literals.

Here's a sample I ran in my browse console to illustrate:

gideonthomas: heh, yep, you're right, I totally read the docs wrong

cadecairos:
@gideonthomas changes made to that destructuring assignment.

PR #1910, opened 2017-11-16
Reverting SEPA string changes

flukeout:
This undoes the string changes we made for SEPA. This relates to the copy underneath the payment options, which should now appear as...

image


flukeout:
cc @gideonthomas - we don't have an issue for this but you asked me to file it.

PR #1911, opened 2017-11-16
Store stripe customer id in cookie

gideonthomas:


server/index.js
```diff
@@ -31,6 +33,12 @@ module.exports = function(options) {
     }
   });
 
+  server.state("stripeCustomerId", {
+    ttl: ONE_HOUR_MS,
+    isSecure: process.env.NODE_ENV === "production",
+    encoding: "base64"
```
cadecairos: Should we use something like [Iron](https://www.npmjs.com/package/iron) to encrypt/decrypt this server-side?

I'd prefer the client to not even be aware of what the customer ID even is, and for it to be difficult to craft a fake cookie.

gideonthomas: sure, I can add that

server/routes/index.js
```diff
@@ -269,7 +271,8 @@ var routes = {
   },
   stripeMonthlyUpsell: function(request, reply) {
     var transaction = request.payload || {};
-    var customerId = transaction.customerId;
+    var cookie = request.yar.get("session");
+    var customerId = cookie && cookie.stripeCustomerId;
```
cadecairos: Shouldn't we return early if this is undefined?

server/routes/index.js
```diff
@@ -282,6 +285,17 @@ var routes = {
       metadata.glassroomnyc = true;
     }
 
+    if (!customerId) {
+      request.log(['error', 'stripe', 'recurring'], {
+        request_id,
+        err: 'Customer ID not present in the cookie'
+      });
+
+      return reply(boom.badRequest('An error occurred while fetching the customer for this monthly upsell'));
```
cadecairos: can we make this error ''An error occurred while creating this monthly donation"

gideonthomas:
OMG so this is ready for review. @cadecairos to actually use a version of `iron` that is compatible with Node 8, I had to use something that works with their async/await structure so I had to add in promise handling -.-

gideonthomas:
OMG @cadecairos wtf, why are you running this on node 8?

gideonthomas:
@cadecairos so if we're updating to node 8 now, let's do that first before landing this...otherwise, I'm going to revert my last two commits and mention that this code needs to be changed in the node upgrade issue.

server/routes/index.js
```diff
@@ -282,56 +319,95 @@ var routes = {
...
-              id: subscription.id
-            }).code(200);
+        stripe.retrieveCustomer(
+          customerId,
+          function(retrieveCustomerErr, customer) {
+            if (retrieveCustomerErr) {
+              return reply(boom.badImplementation('An error occurred while creating this monthly donation', retrieveCustomerErr))
+                .unstate("session");
+            }
+            // Make this with a monthly delay for the user.
```
ScottDowne: I meant to remove this. I'll remove it in another patch I think.

ScottDowne:
I think this looks good to me. Let's get it on the main branch now, to do a final round of testing, if that's the idea.

ScottDowne:
@cadecairos I think I'll be fixing that in a front end patch once this gets merged into the upsell branch.

PR #1916, opened 2017-11-20
Update eslint

gideonthomas:
The code changes are mainly just indent fixes done automatically with `--fix`

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "css-loader": "0.23.1",
     "enhanced-require": "0.5.0-beta6",
     "enzyme": "2.7.0",
-    "eslint": "1.10.3",
+    "eslint": "^4.11.0",
     "eslint-loader": "1.1.1",
     "eslint-plugin-react": "3.14.0",
     "exports-loader": "0.6.2",
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "react-redux": "4.4.5",
     "react-router": "2.0.0",
     "redux": "3.6.0",
-    "redux-persist": "^4.10.1",
+    "redux-persist": "4.10.1",
     "request": "2.67.0",
     "scooter": "2.0.1",
     "stripe": "5.1.1",
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "react-redux": "4.4.5",
     "react-router": "2.0.0",
     "redux": "3.6.0",
-    "redux-persist": "4.10.1",
+    "redux-persist": "^4.10.1",
     "request": "2.67.0",
     "scooter": "2.0.1",
     "stripe": "5.1.1",
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "css-loader": "0.23.1",
     "enhanced-require": "0.5.0-beta6",
     "enzyme": "2.7.0",
-    "eslint": "^4.11.0",
+    "eslint": "4.11.0",
     "eslint-loader": "1.1.1",
     "eslint-plugin-react": "3.14.0",
     "exports-loader": "0.6.2",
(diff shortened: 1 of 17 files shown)


src/components/amount-buttons.js
```diff
@@ -85,13 +85,13 @@ var AmountOtherButton = React.createClass({
     var inputValue = e.currentTarget.value;
     var amount = "";
 
-    if (/^[\d]*[\.]?\d{0,2}$/.test(inputValue)) {
+    if (/^[\d]*[.]?\d{0,2}$/.test(inputValue)) {
```
gideonthomas: actual code change

src/components/amount-buttons.js
```diff
@@ -85,13 +85,13 @@ var AmountOtherButton = React.createClass({
...
       amount = inputValue.replace(/,/g, "");
     } else if (/^[\d]*[,]?\d{0,2}$/.test(inputValue)) {
       amount = inputValue.replace(/\./g, "").replace(",", ".");
-    } else if (/^[\d,]*[\.]?\d{0,2}$/.test(inputValue)) {
+    } else if (/^[\d,]*[.]?\d{0,2}$/.test(inputValue)) {
```
gideonthomas: actual code change

src/components/amount-buttons.js
```diff
@@ -85,13 +85,13 @@ var AmountOtherButton = React.createClass({
...
       amount = inputValue.replace(/,/g, "");
-    } else if (/^[\d\.]*[,]?\d{0,2}$/.test(inputValue)) {
+    } else if (/^[\d.]*[,]?\d{0,2}$/.test(inputValue)) {
```
gideonthomas: actual code change

src/lib/analytics.js
```diff
@@ -11,7 +11,7 @@ module.exports = function() {
 
   // Search for PayPal Params
   var pp_tx_re = /tx=(\w+)/;
-  var pp_amt_re = /amt=([\w\.]+)/;
+  var pp_amt_re = /amt=([\w.]+)/;
```
gideonthomas: actual code change

src/lib/analytics.js
```diff
@@ -31,7 +31,7 @@ module.exports = function() {
 
   // Search for Coinbase Params
   var cb_tx_re = /order\[transaction\]\[id\]=(\w+)/;
-  var cb_amt_re = /order\[total_native\]\[cents\]=([\w\.]+)/;
+  var cb_amt_re = /order\[total_native\]\[cents\]=([\w.]+)/;
```
gideonthomas: actual code change

src/lib/analytics.js
```diff
@@ -42,7 +42,7 @@ module.exports = function() {
 
   // Search for Stripe Params
   var str_tx_re = /str_id=(\w+)/;
-  var str_amt_re = /str_amount=([\w\.]+)/;
+  var str_amt_re = /str_amount=([\w.]+)/;
```
gideonthomas: actual code change

src/components/amount-buttons.js
Pomax: Can you explain the code change? The origin RE matches an optional `\` or `.` in the second character class, was this a bug?

gideonthomas: don't think the original RE matches the optional `\`. I think for it to do that it would need to be `\\` since `\` is an escape character inside the `[ ]`. So theoretically, both before and after this change, the regex should do the same thing. I'm checking in with @ScottDowne though.

package.json
```diff
@@ -56,7 +56,7 @@
     "css-loader": "0.23.1",
     "enhanced-require": "0.5.0-beta6",
     "enzyme": "2.7.0",
-    "eslint": "1.10.3",
+    "eslint": "^4.11.0",
```
Pomax: do we want `^` or do we want to fix the version number.

gideonthomas:
@Pomax r+ via slack

PR #1918, opened 2017-11-20
Fixing a number formatting bug.

ScottDowne:


ScottDowne:
@gideonthomas So, I broke this earlier in the year.

STRS:
1. Go to main donate page
2. Enter a large number, like 1,234.12 to trigger number formatting and with enough complexity to have both comma and period.
3. now add in another number to the end, like 1,234.123

Expected: The number is incorrect, and react shouldn't even let it be inputted.
Actual: clears your input.

If you're wondering why bother, not many people donate thousands of dollars, just remember other currencies are bigger.

What's happening, If the input was invalid, we shouldn't change the value, thus shouldn't update it.
When I broke this, I think I was just looking at it as a refactor, and not thinking of all the fringe cases. In this case, we init the local amount variable with an empty string. So as soon as we hit an invalid input, we end up clearing the input by setting it to the init value when we shouldn't have. I can explain this further if you want, just ping me.

PR #1922, opened 2017-11-24
Add cookie password env to app.json

gideonthomas:
aForgot to update this when I added a new env var

### DIFF
--- a/app.json
+++ b/app.json
@@ -88,7 +88,7 @@
       "required": true
     },
     "SECRET_COOKIE_PASSWORD": {
-      "required": true
+      "generator": "secret"
     },
     "STRIPE_PUBLIC_KEY": {
       "required": true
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -87,6 +87,9 @@
     "SIGNUP": {
       "required": true
     },
+    "SECRET_COOKIE_PASSWORD": {
+      "required": true
```
cadecairos: lets add `"generator": "secret"` to this, it doesn't need to inherit this secret from the staging environment.

cadecairos: Are there other places we can us the generator, so that we're not re-using staging secrets?

gideonthomas: good catch!

gideonthomas: don't think there are other instances

PR #1923, opened 2017-11-24
Updated overflow property for modal.

flukeout:
This way, if the window is too small, the modal will scroll. Mistakenly used `hidden` before. 

cc @ScottDowne 

On a small window, should appear as...

image


ScottDowne:
R+

PR #1927, opened 2017-11-28
Events for monthly upgrade clicks.

ScottDowne:


src/components/monthly-upgrade.js
```diff
@@ -41,6 +46,12 @@ var MonthlyUpgrade = React.createClass({
       return;
     }
 
+    reactGA.event({
```
alanmoo: This fires after the submission goes through completely?

alanmoo: That's based on the check for `state.submitting`...I'm trying to think through if there's any opportunities for this to give us bad data.

ScottDowne: No, on the button click.

ScottDowne: I think, let me double check. On click was my intent.

ScottDowne: Yeah, it's onClick

PR #1929, opened 2017-12-01
Update share images/copy in OG tags on donate.mozilla.org

gvn:
Closes https://github.com/MozillaFoundation/Advocacy/issues/1108


locales/en-US/messages.properties
```diff
@@ -99,7 +99,7 @@ share=Share
 tell_your_friends=Tell your friends!
 help_spread_the_word=We need your help to spread the word.
 # Used in a tweet, please keep below 116 characters
-i_donated_to_mozilla=I donated to @mozilla today because I #lovetheweb. Join me and help build + protect the Web forever
```
TheoChevalier: Please update the key here and in the code, otherwise it won’t be translated.
We also need to update the comment, since Twitter limit has changed. I think we can go with 257 characters now

gvn: What should the key be changed to? Also, what should the copy be?

TheoChevalier: Copy is correct, as far as I can tell. For key/comment, we could go with the following:

Updated key example: `i_donated_to_mozilla_2017`
Updated comment: `# Used in a tweet, please keep below 257 characters`

src/lib/react-server-route.js
```diff
@@ -30,12 +30,12 @@ function routeFileContent(locales) {
     function generateHTML(renderProps) {
       var messages = getMessages(locale);
       var favicon = "/assets/images/favicon.d0608f227db61f2852a32087e614911c.png";
-      var twitterImage = "/assets/images/EOY_Twitter_v8_EN.d1bb5d2a5ce35859d038df852d9e6a0a811beaac.png";
-      var facebookImage = "/assets/images/EOY_facebook_v1.a152496406bad899d1a920f6d6b9f507.png";
```
ScottDowne: Do these images still exist in the repo and if so, can we delete them or is there anywhere we're using them still?

TheoChevalier:
The fixes look good to me 👍

ScottDowne:
@gvn this looks good to me too.

PR #1931, opened 2017-12-05
Donate video tracking events.

flukeout:
I've added a couple of events for tracking the video

* Initial Video Start
  * Keeps track of the first time someone plays the video
* Video Watched Progress
  * Keeps track of how much of the video people watch, in 10% increments

I've tested these by monitoring the network activity and it seems to be sending the correct things in both Firefox and Chrome. Apparently once it's in staging, we can test if it's hitting our analytics account properly.

cc @vojtechmzl


flukeout:
_Update_: Fixing lints now.

src/pages/about.js
```diff
@@ -70,7 +114,7 @@ module.exports = React.createClass({
         <div className="content-wrapper">
           <img className="mozilla-watermark" alt="Mozilla Logo" src="/assets/images/mozilla.1068965acefde994a71c187d253aca2b.svg"/>
         </div>
-        <video controls poster="/assets/images/donate-video-poster.jpg">
+        <video ref="videoPlayer" controls poster="/assets/images/donate-video-poster.jpg">
```
ScottDowne: There is an updated way of storing and referencing refs.

`ref={(videoPlayer) => { this.videoPlayer = videoPlayer; }}`

ScottDowne: Bit more info here: https://reactjs.org/docs/refs-and-the-dom.html

You also reference this with `this.videoPlayer` instead of `this.refs.videoPlayer`

src/pages/about.js
```diff
@@ -48,9 +49,52 @@ module.exports = React.createClass({
...
+    var percentIncrement = 10; // We'll track the video progress in 10% increments.
+    if (this.props.test === "video") {
+
+      this.setState({
+        nextPercentDone: percentIncrement,
+        percentIncrement : percentIncrement,
+        videoStarted : false
+      });
+
+      this.refs.videoPlayer.addEventListener("play", this.onVideoStart);
```
gideonthomas: you have to use `e => this.onVideoStart(e)` here otherwise you'll lose `this` inside the function (same for the line below).

src/pages/about.js
```diff
@@ -48,9 +49,52 @@ module.exports = React.createClass({
...
+      reactGA.event({
+        category: "Video",
+        action: "Video Progress",
+        label: "Donate Video",
+        value: nextPercentDone
+      });
+    }
+  },
+  onVideoStart: function(e) {
+    if (this.state.videoStarted === false) {
```
ScottDowne: Instead of having a state variable, you can remove the onVideoStart event on play.

`this.refs.videoPlayer.removeEventListener("play", this.onVideoStart);`

src/pages/about.js
flukeout: Happy to change, but.. it works!

src/pages/about.js
```diff
@@ -48,9 +49,52 @@ module.exports = React.createClass({
...
+    if (this.props.test === "video") {
+
+      this.setState({
+        nextPercentDone: percentIncrement,
+        percentIncrement : percentIncrement,
+        videoStarted : false
+      });
+
+      this.refs.videoPlayer.addEventListener("play", this.onVideoStart);
+      this.refs.videoPlayer.addEventListener("timeupdate", this.onVideoTimeupdate);
```
ScottDowne: Any events you add, you need to ensure you remove inside a `componentWillUnmount` function. Example https://github.com/mozilla/advocacy.mozilla.org/blob/251c8d1429e637998af43464b817087e5c89c45f/src/components/imagetag.js#L28-L34

src/pages/about.js
gideonthomas: react might be auto-binding some of these functions for you but I wouldn't risk it...any update to react will break this

flukeout: For sure, will update!

src/pages/about.js
ScottDowne: @flukeout not sure if you missed this, or just didn't think it would be worth changing. Both are valid reasons :)

If you think it's not worth it, I'm totally fine with keeping it as a state variable.

flukeout: Noticed, but opting to keep how it is as a state variable, although I like your solution more.

ScottDowne: K, that's fine.

ScottDowne:
Looks good now, just waiting on travis.

flukeout:
Travis says 👍 so I'll merge.

PR #1932, opened 2017-12-07
Initial removal of redux persist.

ScottDowne:
## What redux persist was fixing

autocomplete is an attribute on input elements that does this "The browser is allowed to automatically complete the value based on values that the user has entered during previous uses" see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete

This autocompleted value is what the user sees, but never gets put into the app state, and thus app state gets out of sync with what the user sees.

This could give users the possibility, if they refresh the page after interacting with it, to donate an amount they didn't expect to, or monthly/one time when they meant to give the other way. So it's pretty important.

## Why we chose redux persist

Redux persist was a quick solution that fixed the problem in one place for all inputs. It did this by putting the state of the app in SessionStorage, which matches what the user sees.

This caused query params values to only work on a new tab, which wasn't great.

## This patch's solution without Redux Persist

We should instead just turn off autocomplete on all inputs.

### DIFF
--- a/src/components/currency-dropdown.js
+++ b/src/components/currency-dropdown.js
@@ -18,7 +18,7 @@ var CurrencyDropdown = React.createClass({
   },
   render: function() {
     return (
-      <select onChange={this.onChange} autocomplete="off" className="currency-dropdown" value={this.props.currency.code}>
+      <select onChange={this.onChange} autoComplete="off" className="currency-dropdown" value={this.props.currency.code}>
         {Object.keys(currencyData).map(function(currency, i) {
           return (
             <option value={currency} key={i}>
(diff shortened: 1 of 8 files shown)


ScottDowne:
I want to block this on this pr https://github.com/mozilla/donate.mozilla.org/pull/1945

ScottDowne:
Not blocking this anymore. It's a different approach now, and not nearly as big, so it doesn't need to be blocked.

src/components/currency-dropdown.js
```diff
@@ -18,7 +18,7 @@ var CurrencyDropdown = React.createClass({
   },
   render: function() {
     return (
-      <select onChange={this.onChange} className="currency-dropdown" value={this.props.currency.code}>
+      <select onChange={this.onChange} autocomplete="off" className="currency-dropdown" value={this.props.currency.code}>
```
cadecairos: `autocomplete` => `autoComplete`

ScottDowne: boooo, good catch.

PR #1933, opened 2017-12-08
Adding a few more copy tests.

ScottDowne:


alanmoo:
The hyphen in test 2 makes it feel broken because it falls on a new line. @WillatMozFdn Can we make that a new sentence or something? https://donate-mozilla-org-us--pr-1933.herokuapp.com/en-US/?test=eoy2017test2

WillatMozFdn:
@alanmoo would this edit work? I just took out the dash and made it a new sentence as you suggested.  "An open Web works when people like you come together to protect it. As a non-profit, we rely on donations to carry out our mission.  When you support Mozilla, you’re fighting for a future better, healthy, internet. So please donate today."

alanmoo:
@WillatMozFdn @vojtechmzl These will be live momentarily using https://donate.mozilla.org/en-US/?test=eoy2017test1 and https://donate.mozilla.org/en-US/?test=eoy2017test2

PR #1934, opened 2017-12-08
Adding payment logo for discover for only usd

ScottDowne:
Closes https://github.com/MozillaFoundation/Advocacy/issues/1113


ScottDowne:
@alanmoo this adds discover image to usd only.

I didn't make the sprite issue worse, just worked around it, because we wanted to fix it at a later date.

PR #1945, opened 2018-01-04
Pre populate user email in signup form, and kill redux persist for si…

ScottDowne:
…gnup form.

### DIFF
--- a/src/components/checkbox.js
+++ b/src/components/checkbox.js
@@ -12,9 +12,17 @@ var Checkbox = React.createClass({
     error: React.PropTypes.string,
     name: React.PropTypes.string
   },
+  componentDidMount: function() {
+    // The browser may have a stored state here.
+    // Ensure we are in sync with what the user sees.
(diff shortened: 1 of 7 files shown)


ScottDowne:
I initially wanted to just pre populate the email input with the user's email, example steps:

1. I make a donatation via Stripe.
2. I'm redirected to the thank you page with a signup request.
expected: I see my email pre populating the email input.
without the fix: the input is blank.

I ran into a snag where redux persist was clobbering my input after the redirect, and clearing it.

This is the patch to sort out redux persist https://github.com/mozilla/donate.mozilla.org/pull/1932 once that lands this would be easier, however, that patch is too large.

I found a way in redux persist to turn it off just for the signup form, and leave it for the rest to be dealt with in the other ticket.

ScottDowne:
Closing this, I'll do it in another patch after the autocomplete stuff in https://github.com/mozilla/donate.mozilla.org/pull/1932 lands.

PR #1947, opened 2018-01-08
No zip

ScottDowne:


ScottDowne:
Landed most of this in https://github.com/mozilla/donate.mozilla.org/pull/1948

Going to close this.

I think there is still some dead code that can be removed, but might as well do it in another patch.

PR #1956, opened 2018-01-31
Update Hapi to v17.2.0

cadecairos:
This PR updates Hapi to v17.2.0, and includes updates to all supporting modules.

Hapi v17.2.0 changed a lot of the APIs we used to depend on, so a lot of refactoring was needed.


package.json
```diff
@@ -89,16 +90,15 @@
...
+    "scooter": "5.0.0",
     "stripe": "5.1.1",
     "style-loader": "0.13.0",
-    "throng": "1.0.1",
+    "throng": "4.0.0",
     "webpack": "1.12.9",
     "whatwg-fetch": "0.10.1"
   },
   "devDependencies": {
-    "aws-sdk": "2.2.26",
```
ScottDowne: Was this dead code and  simply not being used or somehow related to this patch?

ScottDowne: Oh I see, it's already listed above.

package.json
```diff
@@ -13,12 +13,11 @@
     "watch:babel": "npm run build:babel -- -w",
     "watch:webpack": "webpack --progress --profile --colors --watch",
     "server": "node start.js",
-    "test": "npm-run-all --parallel test:server test:http test:depcheck test:lint test:karma",
+    "test": "npm-run-all --parallel test:http test:depcheck test:lint test:karma",
     "test:karma": "karma start",
     "test:coverage": "KARMA_ENV=coverage karma start --reporters coverage,coveralls",
     "watch:test": "karma start --no-single-run --browsers Firefox",
-    "test:http": "mocha tests/http/*.test.js",
-    "test:server": "node tests/start.js",
```
ScottDowne: Thoughts on removing this script?

server/base-routes.js
```diff
@@ -154,7 +154,7 @@ var baseRoutes = [
   }, {
     method: 'GET',
     path: '/api/polyfill.js',
-    handler: function(request, reply) {
+    handler: function(request, h) {
```
ScottDowne: What's h?

server/base-routes.js
```diff
@@ -164,15 +164,18 @@ var baseRoutes = [
...
       var params = {
         features: polyfills.get(),
         minify: true,
         unknown: 'polyfill'
       };
       params.uaString = request.plugins.scooter.source;
-      polyfillio.getPolyfillString(params).then(function(bundleString) {
-        reply(bundleString).type('application/javascript; charset=utf-8').vary('User-Agent');
+
+      return new Promise((resolve) => {
```
ScottDowne: I'm guessing new hapi just expects things to be handled as promises now?

server/base-routes.js
```diff
@@ -184,7 +187,7 @@ var baseRoutes = [
   }, {
     method: 'GET',
     path: '/api/client-env.js',
-    handler: function(request, reply) {
+    handler: function(request, h) {
```
ScottDowne: I see h is being used differently in this handler compared to the last.

server/finalise-server.js
```diff
@@ -1,15 +1,11 @@
-var Path = require('path');
-
-module.exports = function finalizeServer(err, server, reactRouted) {
-  if (err) {
```
ScottDowne: This err stuff no longer serving any functions?

server/get-server-options.js
```diff
@@ -2,21 +2,21 @@ module.exports = function getServerOptions(options) {
...
-            preload: true
-          },
-          xframe: process.env.ENABLE_XFRAMEOPTIONS === 'true',
-          xss: true,
-          noOpen: true,
-          noSniff: true
-        }
+    address: process.env.HOST,
+    port: process.env.PORT,
+    uri: process.env.APPLICATION_URI,
```
ScottDowne: These above three are new data. Not that there is another wrong, but is it something required now or is it new data we need access to later in the app?

server/index.js
```diff
@@ -1,35 +1,27 @@
...
-  var server = new Hapi.Server(serverOptions);
-
-  server.connection({
-    host: process.env.HOST,
-    port: process.env.PORT,
-    uri: process.env.APPLICATION_URI
-  });
-
-  server.register(require("hapi-auth-bearer-token"), function(err) {
-    if (err) {
```
ScottDowne: This error check serving no purpose?

package.json
cadecairos: It literally did nothing, AFAICT started up, and turned off the server. we can leave that testing to the Hapi test suite.

server/routes/index.js
```diff
@@ -1,11 +1,10 @@
 const iron = require('iron');
-var signup = require('./signup');
```
ScottDowne: signup being removed?

server/base-routes.js
cadecairos: Hapi's new request/response lifecycle uses this thing they call the "toolkit" and it's shortened to `h` 

A full porting guide is available here: https://github.com/hapijs/hapi/issues/3658 if you're curious

server/base-routes.js
cadecairos: yeah, if it can't return a value synchronously, you return a promise for the value, or use async/await (so... promises, basically)

server/base-routes.js
cadecairos: I could "promisify" that pollyfillio function above, but I chose instead to wrap it in a promise, since I don't really trust `util.promisify` that much. In this case, the request response is available to return without needing to await an async result, so we can just return it.

package.json
ScottDowne: iirc it made it so we would catch it sooner? Best reason I could come up with.

tests/http/static-files.test.js
```diff
@@ -11,233 +11,249 @@ var types = {
...
+    instance = await server();
+    done();
+  });
+
+  Object.keys(files).forEach((url) => {
+    it(`should return okay when accessing ${url}`, async(done) => {
+      let response = await instance.inject({ url });
+      let { type, cache, vary } = files[url];
 
-  var instance = server({ useDomains: false });
```
ScottDowne: useDomains: false is a thing we don't need anymore?

server/finalise-server.js
cadecairos: If something goes wrong in Hapi now, it'll get thrown, and then caught

server/get-server-options.js
cadecairos: Hapi 17 deprecated the "Connection" API, which we passed this information to in server/index.js. You now set up this stuff in the main server config.

tests/http/static-files.test.js
cadecairos: Deprecated feature

package.json
cadecairos: Our test suite starts a server before each new case, so having this test is pretty redundant still

server/routes/index.js
cadecairos: It's required inline on line 71 now

server/index.js
cadecairos: register no longer has a callback api and throws errors. they'll be caught in the Try/Catch wrapping any register calls we have

ScottDowne: k

server/base-routes.js
ScottDowne: ooook.

ScottDowne:
@cadecairos R+ let's try this on staging.

cadecairos:
Let's do it!!

PR #1961, opened 2018-02-13
Fix the disputed charge webhook

cadecairos:
Reverse the incorrect logic that was supposed to determine if a dispute should be closed.


ScottDowne:
Once travis re runs and passes, you can get an r+

PR #1969, opened 2018-03-08
Capture the donation_url in subscription metadata, disable url captur…

cadecairos:
…e for paypal

### DIFF
--- a/server/routes/index.js
+++ b/server/routes/index.js
@@ -460,27 +460,12 @@ const routes = {
       paypal_request_sale_service
     });
 
-    const cookie = { donation_url };
     const response = {
       endpoint: process.env.PAYPAL_ENDPOINT,
       token: checkoutDetails.TOKEN
     };
 
-    try {
-      const encryptedCookie = await encrypt(cookie);
-      return h.response(response)
-        .state("session", encryptedCookie)
-        .code(200);
-
-    } catch (err) {
-      request.log(['error', 'paypal', 'cookie'], {
-        request_id,
-        code: err.code,
-        message: err.message
-      });
-
-      return h.response(response).code(200);
-    }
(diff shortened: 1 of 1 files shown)


server/routes/index.js
```diff
@@ -487,20 +488,6 @@ const routes = {
       locale = '/' + locale;
     }
 
-    let donation_url;
-
-    // don't fail if the donation_url isn't available,
-    // in case someone's browser isn't sending cookies
-    const encryptedCookie = request.state && request.state.session;
```
cadecairos: Discovered on staging that browsers aren't forwarding the Cookie when paypal redirects to the callback. we'll need a different solution.

server/routes/index.js
```diff
@@ -235,6 +235,8 @@ const routes = {
       let stripe_create_subscription_service;
       let subscription;
 
+      metadata.donation_url = donation_url;
```
cadecairos: subscription charges get sent to basket async, bia webhooks, so I'm going to dump the url into the charge metadata, so the webhook can extract  it.

ScottDowne: Is metadata a paypal thing?

ScottDowne: wait, we're in stripe. Not what I thought this was. Or I'm confused.

Github really makes it hard to follow this with expanding in the pr.

server/routes/index.js
```diff
@@ -421,7 +423,6 @@ const routes = {
     let frequency = transaction.frequency || "";
     let currency = transaction.currency;
     let amount = amountModifier.paypal(transaction.amount, currency);
-    let { donation_url } = transaction;
```
ScottDowne: NOW we're in paypal.

PR #1972, opened 2018-03-13
Adding recaptcha front end.

ScottDowne:
WIP I left some notes for things we can still fix.

I havn't done back end yet.

I need to add stuff so it doesn't bork via review app or tests. Basically, turn it on and off when it's production or explicitly turned on locally to test.


src/mixins/stripe.js
```diff
@@ -64,6 +64,14 @@ var StripeMixin = {
       label: error
     });
   },
+  reCaptcha: function(callback) {
+    grecaptcha.render('g-recaptcha', {
```
ScottDowne: So thinking about putting the content of this function inside of some check to make sure `grecaptcha` is on the window, this way we can lazy load the captcha script while the user interacts with the form itself, likely being ready by the time they are done.

src/mixins/stripe.js
```diff
@@ -64,6 +64,14 @@ var StripeMixin = {
...
+      'sitekey' : '6LeVaUwUAAAAAM6TtwWfLQLXEWY5sSzEYdWT3fYK',
```
ScottDowne: Make this an .env

server/routes/index.js
```diff
@@ -70,6 +76,32 @@ const mailchimp = async function(request, h) {
 const routes = {
   signup: require('./signup'),
   mailchimp,
+  'reCaptcha': async function(request, h) {
+    const payload = request.payload || {};
+    const reCaptchaToken = payload.reCaptchaToken || "";
```
cadecairos: with Joi requiring this to be set, does it make sense to even have the `|| "";`

src/mixins/stripe.js
cadecairos: would that be difficult?

server/routes/index.js
ScottDowne: Probably not. I guess if the joi stuff ever gets refactored, this'll fail a bit more gracefully?

app.json
```diff
@@ -84,6 +84,12 @@
     "PAYPAL_USER": {
       "required": true
     },
+    "RECAPTCHA_SECRET_KEY": {
+      "required": true
```
cadecairos: Lets make these the test keys

ScottDowne:
Found a bug while banging against this. If you close stripe an reopen it, recaptcha fails saying it's already been initialized.

server/routes/index.js
```diff
@@ -81,6 +80,9 @@ const routes = {
     const reCaptchaToken = payload.reCaptchaToken || "";
 
     return new Promise((resolve, reject) => {
+      if (process.env.RECAPTCHA_DISABLED) {
+        return resolve("OK");
```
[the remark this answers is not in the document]
ScottDowne: Good idea.

ScottDowne:
Getting this on staging now :rocket:

PR #1988, opened 2018-04-21
Updating messages

ScottDowne:


src/components/single-form.js
```diff
@@ -32,8 +32,12 @@ var singleForm = React.createClass({
     };
   },
   renderPrivacyPolicy: function() {
+    var privicyPolicyMessage = "privacy_policy_var_b";
```
TheoChevalier: typo, privicy (and below)

ScottDowne:
@TheoChevalier Good catch, updated.

ScottDowne:
@TheoChevalier any thoughts on l10n for this and timing?

ScottDowne:
@TheoChevalier just in case I put this behind a lang check, mainly because we're landing captcha this week and I was worried about prod potentially being blocked while we sort out any unexpected stuff with that, so I didn't feel comfortable letting in l10n translations on this, depending on how captcha goes, so just to be safe, I put this behind the lang check so we can potentially land it before captcha for at least english.

TheoChevalier:
@ScottDowne Ok, thanks! Feel free to merge whenever it’s ready (no need to coordinate with me here, as I don’t have to send a heads up)

PR #1991, opened 2018-04-30
Removing bitcoin for now

ScottDowne:


src/components/small-print.js
```diff
@@ -20,10 +20,9 @@ var Footer = React.createClass({
       <div className="row disclaimers">
         <p className="other-ways-to-give">
           <FormattedMessage
-            id='other_way_to_give_wire'
+            id='other_way_to_give_wire_check'
```
alanmoo: Why'd this change?

ScottDowne: The string for that expects bitcoin, and if we just removed the link we would get "Other ways to give: Wire Transfer or SEPA | Check | " Notice the extra " | " at the end.

It happens thunderbird already doesn't do bitcoin, so we have a localized string to display bitcoin and not to display bitcoin, so I just updated it to use the same strings as thunderbird, which is exactly the same, cept for one less link, so we see "Other ways to give: Wire Transfer or SEPA | Check" instead of "Other ways to give: Wire Transfer or SEPA | Check | "

alanmoo: Ah, I did a command F but didn't see that string in the files in this commit.

PR #1995, opened 2018-05-11
Adding traffic cop

ScottDowne:
Don't merge until the removal of optimizely patch lands.


ScottDowne:
@cadecairos @alanmoo this is what I ended up doing to get traffic cop working.

I also cleaned up our ga script to avoid unsafe-inline the quickest way I knew how.

assets/js/ga.js
```diff
@@ -0,0 +1,6 @@
+(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
```
alanmoo: As much as it pains me, we should be wrapping this in Do Not Track, shouldn't we @cadecairos?

alanmoo:
Now that the optimizely removal has landed, can we get this moving forward? CI Fixed and then a review?

assets/js/ga.js
ScottDowne: Yeah totally, we can probably just file it.

ScottDowne:
@alanmoo travis gives off a lot of random false positives for donate, so in most cases it's fine and you just need to tell travis to rebuild until it run properly.

ScottDowne:
@alanmoo we should get this on staging today. I think it's ready to go, and travis had a complete run.

assets/js/ga.js
```diff
@@ -0,0 +1,6 @@
...
+(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+ga('create', 'UA-49796218-32', 'auto');
```
Pomax: noscript, privacy badger, etc., even a "defer all scripts" addon can make this fail. Always a good idea to wrap this in `if (typeof ga === "function") { ... }` because this code should not lead to an error in the console in production.

assets/js/ab-tests.js
```diff
@@ -0,0 +1,10 @@
+// Test example
```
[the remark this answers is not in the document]
alanmoo: Assuming I could just uncomment this and it would work as a test, I'm a fan of leaving this in. It shows where the tests live, and doesn't send me off on a goose chase if I'm not familiar with the A/B testing setup. It might benefit from a link to those docs for further reading though.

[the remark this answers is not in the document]
Pomax: it's the "I can uncomment this and code that _not our code right now will run_" part that I'm apprehensive about =)

At the very least, let's change this commented code so it has no _real_ effect even when uncommented.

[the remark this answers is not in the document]
ScottDowne: I'm just going to link to the docs if that's all the same. I agree with Alan that having an example there is a nice convenience, but if it goes against other principles, having an example that has no effect isn't as helpful. I would rather just a link at that point.

Also, once we have had a test history, I dunno if any of this matters as much.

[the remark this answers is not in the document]
ScottDowne: I'm adding some more info to our readme and docs.

assets/js/ga.js
```diff
@@ -0,0 +1,33 @@
+var _dntStatus = navigator.doNotTrack || navigator.msDoNotTrack;
+
+var fxMatch = navigator.userAgent.match(/Firefox\/(\d+)/);
+
+var ie10Match = navigator.userAgent.match(/MSIE 10/i);
+
+var w8Match = navigator.appVersion.match(/Windows NT 6.2/);
+
+if (fxMatch && Number(fxMatch[1]) < 32) {
```
Pomax: Hm, is there a reason that `var _dntStatus = navigator.doNotTrack || navigator.msDoNotTrack || "Unspecified"` is not good enough? Testing FF version numbers or even testing for a version of IE that was fully discontinued in 2016 seems like we're shimming too much. If the browser's been configured to lie in its `navigator.doNotTrack || navigator.msDoNotTrack` response, then the user clearly messed with their browser and we don't need special considerations?

cadecairos: Lets drop the extra checks, and only look for `var _dntStatus = navigator.doNotTrack || navigator.msDoNotTrack || "Unspecified"` as @Pomax has suggested

assets/js/ga.js
```diff
@@ -0,0 +1,14 @@
+var _dntStatus = navigator.doNotTrack || navigator.msDoNotTrack;
```
Pomax: This is over-engineering the check: the idea is that we want the next `if` block to kick in if DNT is not enabled, so based on [the MDN docs for DNT](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack) we can achieve that with with a direct check:
```
const doNotTrack = navigator.doNotTrack || navigator.msDoNotTrack || window.doNotTrack;
if (!doNotTrack || doNotTrack === 'no') {
  ...
}
```
(note `window` rather than `navigator` is used by IE11, Edge, and Safari 7.1.3+... yay standards!)

alanmoo:
@cadecairos can you update your original so it's using the latest version of this code? (Or is there somewhere else we can put this to re-use it? I can't tell if an npm module is overkill)

cadecairos:
making it an npm module feels like a step too far, But I can revise the original Gist

PR #1996, opened 2018-05-14
Clean up strings, add alternative subject line

TheoChevalier:


TheoChevalier:
Just realized more strings got removed with the Thank You page removal in https://github.com/mozilla/donate.mozilla.org/commit/5f3ae4b606812b66488dfcda785690c2c3f7deda#diff-6027d93aa4254ec29cf42a213c11bbc8 , so removed them from the .properties as well

PR #2000, opened 2018-05-29
Switch On new legal notice for all locales

TheoChevalier:
@ScottDowne We got the Spanish translations a few days ago, we’re now good to expose the string to everyone.
Can’t test this patch locally right now, I need to fix my config, but should be safe… Mind merging if this works for you?


ScottDowne:
@TheoChevalier Yup, that works.

PR #2003, opened 2018-06-04
Changed the heart fill color to red

WesleyBranton:
The heart image looks a little boring with the plain white color, so I think it should be changed to red.


alanmoo:
@kristinashu any thoughts on this?

WesleyBranton:
@kristinashu am I missing something? There is only one heart on the donate.mozilla.org page and the site is quite bland looking. Am I not looking at the correct site?

kristinashu:
Oh sorry @WesleyBranton! I thought this was for a different project. I still think we should keep the heart as is though since we don't want to distract users from the main focus of donating. 

We are currently exploring new designs for this page and plan to try out different treatments so we'll keep this in mind!

WesleyBranton:
@kristinashu is there a link to a discussion about the redesign or anything? I'd love to submit my ideas.

WesleyBranton:
Thanks

PR #2004, opened 2018-06-04
re add thank you links.

ScottDowne:


ScottDowne:
@TheoChevalier you can use the heroku app, examples: 

https://donate-mozilla-org-us--pr-2004.herokuapp.com/de/thank-you-superhero/
https://donate-mozilla-org-us--pr-2004.herokuapp.com/en-US/thank-you-superhero/

PR #2007, opened 2018-06-05
Legal notice string is missing in the FAQ

TheoChevalier:
@ScottDowne Mind giving it a quick look and deploy this one when the pipes are green?

It’s fixing this on both FAQs
image

The string we changed for the legal notice in the footer was actually used in more places, and we forgot to update them


ScottDowne:
R+

PR #2013, opened 2018-06-07
Adding subscribed to paypal.

ScottDowne:


server/base-routes.js
```diff
@@ -143,7 +143,8 @@ const baseRoutes = [
           locale: Joi.string().min(2).max(12).required(),
           currency: Joi.any().valid(currencyFor.paypal).required(),
           appName: Joi.string(),
-          donation_url: Joi.string().required()
+          donation_url: Joi.string().required(),
+          subscribed: Joi.string().min(1).max(1)
```
cadecairos: If this is going to be `0` or `1`only, lets make this `Joi.string().valid('0', '1')`

ScottDowne: ah that's better. Min max doesn't really work with a string anyway? Good catch.

PR #2018, opened 2018-06-21
Fixing watch on js and css

ScottDowne:


ScottDowne:
@cadecairos Should do it. See these docs https://hapijs.com/tutorials/serving-files

Issue #2008, opened 2018-06-05
[Donate] Implement modern reskin


alanmoo:
Placeholder for work to be described in https://github.com/mozilla/foundation.mozilla.org/issues/1534

@kristinashu Please update this comment when everything is ready.

kristinashu:
Based on design in https://github.com/mozilla/foundation.mozilla.org/issues/1534

All comps are here: https://redpen.io/p/thbfb2535f83162d2d
- Colors and font styling is from https://foundation.mozilla.org/style-guide/
- Icons are from current https://donate.mozilla.org/en-US/
- Nav should function like and link to foundation site [zenmode style](https://foundation.mozilla.org/campaigns/facebook-update/) (footer should be the same too)
- Redlines for desktop and hover states
![donate x-large 2x](https://user-images.githubusercontent.com/1068783/41006536-ca16005a-68d6-11e8-84cf-8a4ad5b32947.jpg)

ScottDowne:
Adding a few more possible thunderbird people @ryanleesipes and @sancus

Thoughts?

ScottDowne:
@kristinashu have the changes to the order of the disclaimers been run by legal? They have been very particular in the past on the positioning of the stripe/paypal bit. I'm also not sure what they'll think about making it lighter in colour.

kristinashu:
Also, I didn't include the reCAPTCHA logo (bottom right) in my comp but we should keep that.

<img width="88" alt="image" src="https://user-images.githubusercontent.com/1068783/41786810-589ef904-75fb-11e8-8523-541c53f834d9.png">

kristinashu:
Updated:

![donate x-large copy 9 2x](https://user-images.githubusercontent.com/1068783/41787538-9327f04c-75fd-11e8-94c5-6072dc950042.jpg)

ScottDowne:
Here is where I am currently https://donate-mozilla-org-us--pr-2020.herokuapp.com/en-US/

ScottDowne:
I figure another day, and another day of qa/reviews?

kristinashu:
Sweet!

alanmoo:
Nice. You’re probably aware, but it definitely needs some mobile love.

ScottDowne:
@kristinashu updates https://donate-mozilla-org-us--pr-2020.herokuapp.com/en-US/

kristinashu:
P1
- Zen mode nav (because without there is no mozilla logo)
[...]

PR #2020, opened 2018-06-22
New styles

ScottDowne:
Fixes: https://github.com/mozilla/donate.mozilla.org/issues/2008


ScottDowne:
It should do an AB test for the nav.

alanmoo:
@gvn Please work with Scott to dig into the why and how of this. I recognize it'll take a while but it's worth getting this right. (It's not on fire though)

ScottDowne:
@gvn yeah, but I think @kristinashu did that in the issue not the pr.

Also good catch on the footer background.

gvn:
Ok, in that case, code looks good!

PR #2021, opened 2018-06-25
Updating polyfill-service

ScottDowne:


cadecairos:
@ScottDowne "22 new vulnerable dependencies" :|

Can we just not use polyfill-service, please?

PR #2027, opened 2018-07-05
Fixing bug with thunderbird small print sections.

ScottDowne:
There are links in the small print section of the donate page that link out to various bits of info, some of these are mozilla specific, and some are thunderbird specific.

Before the reskin, the small print lived outside the generic form component and inside the pages themselves, which were specific to thunderbird or mozilla, so it was easy to display the correct small print bits.

When we did the reskin and moved this small print section into the form component I forgot to add a thunderbird version. So links in the small print for thunderbird started to link to mozilla specific stuff.

STRs:
1. go to donate.mozilla.org/thunderbird
2. click on any of the links in the disclaimer that are thunderbird specific, example: either of these "Wire Transfer or SEPA | Check"

Expected: should go to articles listed on the thunderbird faq
Actual: goes to articles on the mozilla donate faq

The fix:

There is already a thunderbird specific small print component to go along with the mozilla one. We need to do 2 things.
1. Update the old thunderbird small print to properly display in the new place.
2. Ensure the form component has the ability to display one or the other.

### DIFF
--- a/src/components/single-form.js
+++ b/src/components/single-form.js
@@ -9,6 +9,7 @@ import { PayPalButton, StripeButton } from './payment-options.js';
 import SubmitButton from './submit-button.js';
 import DonateButton from './donate-button.js';
 import SmallPrint from './small-print.js';
+import SmallPrintThunderbird from './thunderbird/small-print.js';
 import currencies from '../data/currencies.js';
 
 import { connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


src/components/single-form.js
```diff
@@ -91,6 +92,16 @@ var singleForm = React.createClass({
 
     return this.renderWithDisabledCurrency(className);
   },
+  renderSmallPrint: function() {
+    if (this.props.appName === "thunderbird") {
```
Pomax: Slightly more DRY:
```
let Type = (this.props.appName === "thunderbird") ? SmallPrintThunderbird : SmallPrint;
return <Type frequency={this.props.frequency/>;
```

ScottDowne:
Updated

Pomax:
Looks like `tough-cookie` is missing from the package list, too.

Issue #2029, opened 2018-07-09
Updates from legal


kristinashu:
Updates from Abigail

(1) https://donate.mozilla.org/en-US/ways-to-give

Add: "Please remember that donations are deductible for U.S. income tax purposes only."
(text can go under the "New! We are now able to accept donations in Euros..." paragraph")

(2) https://donate.mozilla.org/en-US/thunderbird/

Add: "While Thunderbird is now an independent project separate from Mozilla, Mozilla has agreed to collect donations on our behalf."
(text can go on the left side, under "We’re the leading open source cross-platform email and calendaring client...,")

Update: "Contributions go to the Mozilla Foundation, a 501(c)(3) organization, to support Mozilla Thunderbird" to say instead: "Contributions go to the Mozilla Foundation, a 501(c)(3) organization based in Mountain View, California, to support Thunderbird. They are tax-deductible in the U.S. to the fullest extent permitted by law."

WillatMozFdn:
That verbage sounds good to me!  adding @TheoChevalier for localization into all other languages that we support on the donation pages (both T-bird and Mozilla).

PR #2031, opened 2018-07-10
Legal updates

ScottDowne:
Fixes: https://github.com/mozilla/donate.mozilla.org/issues/2029


locales/en-US/messages.properties
```diff
@@ -48,6 +48,7 @@ problems_donating_thunderbird2=Problems donating? <a href="https://donate.mozill
 donation_notice_2=Contributions go to the Mozilla Foundation, a 501(c)(3) organization based in Mountain View, California, to be used in its discretion for its charitable purposes. They are tax-deductible in the U.S. to the fullest extent permitted by law.
 
 donation_notice_thunderbird=Contributions go to the Mozilla Foundation, a 501(c)(3) organization, to support Mozilla Thunderbird.
```
TheoChevalier: could you remove this string? In general if we keep a string that’s currently not used, I try to add`# Obsolete string` as a localization note so that we don’t forget to remove it later. But here I think we should just remove it

ScottDowne:
Updated

Issue #2026, opened 2018-07-03
Fonts on newly re-skinned page


kristinashu:
Follow up from https://github.com/mozilla/donate.mozilla.org/issues/2008

Looks like prod isn't using the foundation [style guide](https://foundation.mozilla.org/style-guide/). Comps https://redpen.io/p/thbfb2535f83162d2d.

Screen cap from prod:
<img width="1397" alt="image" src="https://user-images.githubusercontent.com/1068783/42247215-970ea02c-7ed4-11e8-8932-9a31d747fa79.png">

cc @ScottDowne

PR #2032, opened 2018-07-12
Zilla slab

ScottDowne:
Fixes https://github.com/mozilla/donate.mozilla.org/issues/2026

### DIFF
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -18,6 +18,8 @@ donate_now=Donate now
 other_amount=Other amount
 one_time=ONE TIME
 monthly=MONTHLY
+one_time_uppercase=ONE TIME
+monthly_uppercase=MONTHLY
 choose_payment=Choose Payment
 secure=Secure
 please_select_an_amount=Please select an amount
@@ -26,6 +28,7 @@ try_again_later=Something seems to have gone wrong. Please try again later
 could_not_complete=The transaction could not be completed. Please email donate@mozilla.org for assistance with your donation.
 other_amount_format_error=Please enter a number up to 2 decimal places.
 donate_button=DONATE
+donate_button_uppercase=DONATE
 credit_card=Credit / debit card
 submitting=Submitting…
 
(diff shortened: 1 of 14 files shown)


ScottDowne:
Should be a review app up soon.

locales/en-US/messages.properties
```diff
@@ -16,16 +16,16 @@ donate_monthly=Donate monthly
 # Homepage
 donate_now=Donate now
 other_amount=Other amount
-one_time=One time
-monthly=Monthly
+one_time=ONE TIME
```
TheoChevalier: Unfortunately this will work only in en-US, we need to retranslate those strings to get an uppercase version.

Since those are critical strings, my suggestion would be to keep this change for en-US **and** add new strings (e.g. `one_time_uppercase`). Then once a critical mass of translations has been reached, swap them. This way we avoid adding more strings behind conditionals

ScottDowne:
I updated it with more changes for content in the footer.

locales/en-US/messages.properties
TheoChevalier: Yep that’s correct, but I’m not sure that’s a en-US only thing. I would say it’s more of a design choice, and if so, it should be applied to all locales.

@kristinashu, do you have an opinion or preference about this?

ScottDowne: I think I keep forgetting the nuances here, but if it's a design choice I would do the all caps in css, the reason I ddon't is because I thought all caps meant different things to different cultures, and thus it was best to let the localizers make the call?

If I'm wrong on this then I can make them all caps in css which would be ideal.

TheoChevalier: No, you’re perfectly right to think that, and that’s why I didn’t suggest a CSS fix. Using `text-transform: uppercase` is a bad practice for localized content https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Create_localizable_UI

ScottDowne: OK great, adding the one_time_uppercase string.

TheoChevalier:
Let’s get this cook on staging for a bit

PR #2055, opened 2018-08-17
Add download link banner to Thunderbird post-download donation page.

Sancus:
Hey,

We'd like to add a banner linking back to the Thunderbird site on the post-download donation page. The reason is because some people are confused by the redirect, or their download doesn't start, making them feel forced to donate or unable to download from the front page.

Here's what that looks like: image

I haven't done much with React before, so let me know if there's a better way to do any of this, I just chose what appeared to be the least invasive way.

### DIFF
--- a/less/pages/thunderbird.less
+++ b/less/pages/thunderbird.less
@@ -1,4 +1,13 @@
 .thunderbird {
+  .download-failed {
+    text-align: center;
+    background: #fff;
+    h3 {
+      font-weight: 700;
+      font-family: "Nunito Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
+    }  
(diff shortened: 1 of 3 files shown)


kristinashu:
Looks good! My only recommendation would be to use the sans serif font Nunito (similar to the "Thank you for downloading Thunderbird" text on the left site) for the text and maybe to move the banner up so it's against the top of the browser.

Here's a quick mockup:
thunder


For reference, we have one like this on the foundation site:
image

Sancus:
I've made the changes requested, fixed the code issues and changed the font and font weight, new screenshot:

image

kristinashu:
Design looks good to me!

PR #2063, opened 2018-08-31
Add a-b-c test for homepage image

### LINKED ISSUE #2045: [Implementation ticket] Test photos on donation landing page
_Tagging @alanmoo for triage_

Based on discussion here: #2033

Here are the implementation specs for the 2 chosen variations:
### Variation 1 (comp):
image

RedPen: https://redpen.io/doa1fcb24c04d2a72f

### Variation 2 (comp):
image

RedPen: https://redpen.io/sz3a8fb434d9b5bbe7

### For both of them, here are the specs:
Spec screenshot: https://redpen.io/bj9f0108a7be1c330b
- [ ] DESKTOP: image has 16:9 proportion and it is the width of the text column. Margin bottom = 32px (same as implemented version)
- [ ] MOBILE: image has a different crop from desktop (smaller height because we don't want to take too much space from the already tiny mobile screen), width is the same as the one from text column, height is proportional as it scales. Margin bottom = 32px (same as implemented version)

### Cropped pictures here (ready to use):
image_i-believe_desktop 2x
image_i-believe_mobile 2x
image_teaching_desktop 2x
image_teaching_mobile 2x

### Notes:
Please connect with @WillatMozFdn to set up the A/B test with him. Also, please keep me in the loop of when it is implemented (so I can do a design review) and when the test launches :). I am curious to see the results of this test.

alanmoo:
Closes #2045 
https://donate-mozilla-org-us--pr-2063.herokuapp.com/en-US/?test=img-a
https://donate-mozilla-org-us--pr-2063.herokuapp.com/en-US/?test=img-b
https://donate-mozilla-org-us--pr-2063.herokuapp.com/en-US/?test=noimg


alanmoo:
@Pomax don't merge this until it also has design approval.

@taisdesouzalessa make sure to test at in-between sizes, I had to make a guess or two.

assets/js/ab-tests.js
```diff
@@ -1,14 +1,13 @@
...
-//   });
-// 
-//   tc.init();
-// }
+var tc = new Mozilla.TrafficCop({
+    id: "image-experiment",
+    variations: {
+      "test=img-a": 33,
+      "test=img-b": 33,
+      "test=no-img": 34
```
Pomax: omg bias~ 

>_>

src/pages/about.js
```diff
@@ -24,10 +24,24 @@ module.exports = React.createClass({
       aboutCopy: aboutCopy
     });
   },
+  imageTest: function() {
+    if (this.props.test === "img-a") {
+      return (<picture>
+        <source srcSet="/assets/images/homepage-images/teacher-wide.jpg" media="(max-width: 800px)" />
+        <img className="homepage-image" src="/assets/images/homepage-images/teacher-full.jpg" alt="Man pointing at student's computer screen"/>
+      </picture> );
+    } else if (this.props.test === "img-b") {
```
Pomax: you don't need `else` if you return in the previous conditional.
```
if (x) { return blah; }
if (y) { return blah2; }
return finalThing
```
Additionally, it'll be good to cache that prop val in a variable
```
let test = this.props.test;
if (test === "img-a") {
   return ...;
}
if (test === "img-b" {
  return ... ;
}
return ...;
```

src/pages/about.js
```diff
@@ -24,10 +24,24 @@ module.exports = React.createClass({
       aboutCopy: aboutCopy
     });
   },
+  imageTest: function() {
+    if (this.props.test === "img-a") {
+      return (<picture>
+        <source srcSet="/assets/images/homepage-images/teacher-wide.jpg" media="(max-width: 800px)" />
+        <img className="homepage-image" src="/assets/images/homepage-images/teacher-full.jpg" alt="Man pointing at student's computer screen"/>
```
Pomax: I notice that this image no longer prespecifies its size, which the original still does. If we can, can we prespecify the dimensions again?

alanmoo: It's not clear the best way to do this with responsive images. Punting for now, as this is a bit of a micro-optimization

alanmoo:
@taisdesouzalessa 

1.) Yeah, review app issue.
2.) Nav is an a/b test on production, which is removed and will always show up going forward, as on staging.
3.) I updated the `picture` element size, so it should work right now.

taisdesouzalessa:
Thanks for the fix and the clarification here @alanmoo. Can't wait to see the results of this test. Are we implementing it as of today? Let me know when it happens so I can annotate on GA :)

Issue #2071, opened 2018-09-20
Thunderbird download banner is broken, "&test=" GET conflict.


Sancus:
It looks like you guys started using "test" for the GET variable for an a-b-c test, unfortunately that's the one that Thunderbird was using for "tbdownload" as well. This results in the donate URL sometimes redirecting to things like `https://donate.mozilla.org/en-US/thunderbird/?utm_source=thunderbird.net&utm_medium=referral&utm_content=post_download&test=tbdownload&test=no-img` which produces a page with no banner at the top. Here was the original PR for TB: https://github.com/mozilla/donate.mozilla.org/pull/2039/files#diff-7018dfc2ff7641275e2f5d96b0788369

If "test" is likely to be used again in this manner it seems most reasonable to just rename the Thunderbird get variable, I can easily change it on our side as well. Otherwise it's likely this will reoccur when someone else forgets.

alanmoo:
Ah, sorry about that! Where does your testing query string get injected? We're running Traffic Cop [here](https://github.com/mozilla/donate.mozilla.org/blob/gh-2066/assets/js/ab-tests.js#L12), so I didn't think to check elsewhere for Thunderbird. (Also, @ScottDowne, who has been more familiar with the donate repo, is no longer at MoFo so apologies for the growing pains)

alanmoo:
After some further digging, we've been running tests using that query param since [June](https://github.com/mozilla/donate.mozilla.org/commit/65843b8a8f4025499b756afb197a1a6cac4e2337). I also can't seem to easily change the name of the parameter for some reason. If it's easy enough for to do so on your side, can you resolve this with that approach for now? Next time we do a test I'll encourage namespacing.

alanmoo:
Another update: The reason I can't change it is #2072, but I unfortunately don't have anyone to work on that right now.

Sancus:
That's OK, it's a [1 line change for us](https://github.com/thundernest/thunderbird-website/blob/master/helper.py#L303), we just set it on download hrefs only on thunderbird.net.

In the donate code, all lines in `/src/pages/thunderbird/about.js` referencing `this.props.test` should be changed. I assume this.props is automatically populated with any get variables? Not super familiar with node/react.

[...]

PR #2078, opened 2018-09-26
Issue #2071: Fix Thunderbird download banner and change query variable to avoid conflicts.

Sancus:
This uses a new query string variable for Thunderbird rather than overloading the 'test' one that is used for other things.

I noticed that these are defined in src/lib/QueryParser.js, but I'm not sure what benefit there is to putting them in there rather than just using this.props.location.query, and that seems to work fine, so I did not modify QueryParser.js so as to not add thunderbird-related clutter that may not be useful to you guys.

If there is a reason that query variables need to be parsed first in that file, let me know and I'll change it.

### DIFF
--- a/src/pages/thunderbird/about.js
+++ b/src/pages/thunderbird/about.js
@@ -15,7 +15,7 @@ module.exports = React.createClass({
 
   componentDidMount: function() {
     var aboutCopy = (<span>{this.context.intl.formatMessage({id: 'additional_info_thunderbird'})}</span>);
-    if (this.props.test === "tbdownload") {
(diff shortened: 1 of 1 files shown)


Sancus:
Yeah, if you can comment on the issue when this is going to go live that would be handy. No need to coordinate any specific time, I'll see it and push the change on our side live, and it's not a big deal if there's a few hours difference, since this is broken on live anyway.

alanmoo:
I'm going to merge this in now and then we'll do a deploy on Monday (cc @TheoChevalier for visibility)

PR #2085, opened 2018-10-03
Remove image b from image tests

### LINKED ISSUE #2084: Take a losing photo out of the donation landing page rotation
The version with "image B" -- the sticker hanging on netting -- is losing in the test.

Please remove from the rotation, and reset the split to go 50-50 between the no-image control version, and image A (guy teaching kids).

Thanks!  @alanmoo

alanmoo:
Closes #2084


src/pages/about.js
```diff
@@ -29,12 +29,6 @@ module.exports = React.createClass({
         <img className="homepage-image" src="/assets/images/homepage-images/teacher-full.jpg" alt="Man pointing at student's computer screen"/>
       </picture> );
     }
-    if (test === "img-b") {
-      return (<picture>
-        <source srcSet="/assets/images/homepage-images/believe-wide.jpg" media="(max-width: 799px)" />
```
mmmavis: Should we remove the 2 image files from repo or is it better that we leave them there in case future A/B testing needs it?

alanmoo: Good call. Done. Let's hope Travis doesn't get upset.

Issue #2074, opened 2018-09-25
More clearly indicate that monthly donations are monthly


alanmoo:
We want to make sure donors are aware that they're donating monthly. Let's add the appropriate indications when the "monthly" radio button is selected.

![image](https://user-images.githubusercontent.com/1682681/46496742-cf5a3080-c7e6-11e8-9cc6-a901f1f6052a.png)
(Add the text inside each of the fields, not just the one)

Make sure to build this such that it can be localized.

alanmoo:
Still some questions about design- this will be updated when ready.

mmmavis:
@alanmoo in terms of text alignment, do you know if we want to

1. center align `$5` with the button and align `per month` to the right of `$5`?
2. OR make `$5` and `per month` one block and center align it?

alanmoo:
2 sounds easier/more flexible. Go with that.

alanmoo:
Good question, Theo. @jessevondoom?

alanmoo:
This landed!

PR #2091, opened 2018-10-05
Indicated monthly donation more clearly

mmmavis:
Related to #2074

### DIFF
--- a/less/components/amount-buttons.less
+++ b/less/components/amount-buttons.less
@@ -29,7 +29,7 @@
     margin-bottom: 10px;
   }
 
-  @currency-symbol-width: 80px;
+  @currency-symbol-width: 70px;
 
   .currency-symbol-container {
     width: @currency-symbol-width;
--- a/less/components/amount-buttons.less
+++ b/less/components/amount-buttons.less
@@ -64,6 +64,17 @@
   .amount-button {
     letter-spacing: -1;
     margin-bottom: 0;
+
+    .per-month-note {
+      display: inline-block;
+      font-size: 11px;
+      margin-left: 5px;
+      width: 33px;
+      color: #ccc;
+      line-height: 1;
+      text-align: left;
+      white-space: normal;
+    }
   }
 
   @media screen and (max-width: 480px) {
--- a/less/components/amount-buttons.less
+++ b/less/components/amount-buttons.less
@@ -29,8 +29,10 @@
     margin-bottom: 10px;
   }
 
+  @currency-symbol-width: 80px;
+
   .currency-symbol-container {
-    width: 60px;
+    width: @currency-symbol-width;
     text-align: center;
     display: block;
   }
@@ -56,7 +58,7 @@
   .amount-other-wrapper {
     position: absolute;
     top: 0;
-    left: 60px;
+    left: @currency-symbol-width;
     right: 0;
     bottom: 0;
   }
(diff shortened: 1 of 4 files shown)


mmmavis:
Not sure if this design would work on mobile though... especially when `per month` gets localized into other languages...

less/components/amount-buttons.less
```diff
@@ -64,6 +64,17 @@
   .amount-button {
     letter-spacing: -1;
     margin-bottom: 0;
+
+    .per-month-note {
+      display: inline-block;
+      font-size: 11px;
+      margin-left: 5px;
+      width: 33px;
```
TheoChevalier: Could this actually be dynamic when there’s room? Many locales will have visual issues, this would help to at least fix the situation on large screens and/or when the amount + currency is short.

locales/en-US/messages.properties
```diff
@@ -15,7 +15,9 @@ donate_monthly=Donate monthly
 
 # Homepage
 donate_now=Donate now
+per_month=per month
 other_amount=Other amount
+other_amount_monthly=Other monthly
```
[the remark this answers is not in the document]
mmmavis: not sure but it sounds weird to me 😬

less/components/amount-buttons.less
mmmavis: Thinking about this more I wonder if we should always put the `per month` note on a separate line (instead of positioning it next to the `[money amount]`). We can't predict how long each localized string and converted currency string are going to be. There might be cases where among the 4 amount buttons we have, few have `[money amount]` and `per month` next to each other and the rest have the 2 strings on separate lines. Design wise this will look a bit odd in my opinion...

mmmavis: tagging @xmatthewx for design feedback - I'm not sure if we wanna get the current design live first and worry about refinement after or ?

xmatthewx: @mmmavis – you're probably right. Putting per month on a second line will make it more flexible for localization. 

I used `display:block; margin: -4px 0 6px;` to tighten in up, but you see what you think is best. 

I wonder if we should handle the Other option in the same way so we can preserve the height (removing monthly from the placeholder and putting in the label).  
 

image

xmatthewx: @alanmoo - feel free to chime in. I'm less familiar with this tool and the original request than you are...

alanmoo: This makes sense to me, still fits the original need.

mmmavis: Thanks! I'll update my PR and flag @TheoChevalier for review again.

locales/en-US/messages.properties
```diff
@@ -15,7 +15,9 @@ donate_monthly=Donate monthly
 
 # Homepage
 donate_now=Donate now
+per_month=per month
```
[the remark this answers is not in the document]
mmmavis: hey @TheoChevalier I updated the PR so it reflects the new design (see screencap below)



(Note I left the existing "Other amount" string untouched 'cuz I wasn't sure if we wanted to change it to `Other` as I assume that new string will have to be localized again?)

I still don't get how we are deal with RTL cases. Do you mind updating my PR to show me how this is usually handled? Thanks!

TheoChevalier:
Should this be merged/deployed, or are we waiting for anything else? @alanmoo

jessevondoom:
Fantastic work everyone! There's nothing in the way of merging/deploying this now.

TheoChevalier:
Thanks Jesse! Merging this now so that people can start translating the new string, and will likely deploy tomorrow morning, before FR + DE fundraisers go out

PR #2119, opened 2018-11-18
Update HTTP to HTTPS

missuniverse:
Changed links to HTTPS when supported.


Pomax:
Please unstage all the locales changes: they get autogenerated by Pontoon, and any changes to them get reset during a locale string update pass.

TheoChevalier:
> Please unstage all the locales changes: they are autogenerated files that we get from Pontoon, so any manual changes to them get reset during a locale string update pass.

Changes to locale files will be properly imported into Pontoon and it’s the correct way of fixing it in this case, but I’m just not sure it’s worth losing translator attributions just for this change. But it’s also just one string…

Issue #2097, opened 2018-10-10
Internet Explorer (11) Not Able to Donate


ryanleesipes:
I have seen quite a few reports of users attempting to donate with IE11 and finding that they are unable to do so.

Tested with Windows 7 and IE11 and unable to set amount or select a payment method.

If IE11 is not supported we should ensure the users know that.

alanmoo:
We should see what kind of lift it is to get this to work. As much as it pains me to support IE11, we shouldn't leave money on the table.

@ScottDowne @cadecairos Any chance you can recall if there were other technical reasons not to support IE11?

ryanleesipes:
Any movement on this or clarification? I continue to get many reports on this front, as @alanmoo pointed out, we're just leaving money on the table.

alanmoo:
This may be a result of the A/B testing using TrafficCop. I disabled it locally and noticed that IE11 makes further progress than previously. If #2104 fixes it, we'll make sure to block IE11 from loading A/B tests in the future.

ryanleesipes:
Thanks for the update and the PR, let's see if it works!

alanmoo:
It seemed to work on staging! Just deployed to production.

ryanleesipes:
Thank you!

PR #2127, opened 2018-12-13
lth and htl a/b testing

Pomax:
bypasses IE11, due to https://github.com/mozilla/donate.mozilla.org/issues/2097

### DIFF
--- a/src/lib/queryParser.js
+++ b/src/lib/queryParser.js
@@ -29,11 +29,11 @@ module.exports = function(queryString, locale) {
     presets = currency.presets[frequency];
   }
 
-  if (test && toString.call(test) === "[object Array]") {
+  if (test && test.join && test.map) {
     test = test.join(" ");
   }
 
-  return {
+  var parsed = {
     values: {
       test: test,
       subscribed: queryString.subscribed
@@ -46,4 +46,6 @@ module.exports = function(queryString, locale) {
       email: queryString.email || ""
     }
   };
+
+  return parsed;
 };
--- a/src/lib/queryParser.js
+++ b/src/lib/queryParser.js
@@ -29,7 +29,8 @@ module.exports = function(queryString, locale) {
     presets = currency.presets[frequency];
   }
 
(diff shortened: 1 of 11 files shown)


src/lib/queryParser.js
```diff
@@ -29,11 +29,11 @@ module.exports = function(queryString, locale) {
     presets = currency.presets[frequency];
   }
 
-  if (test && toString.call(test) === "[object Array]") {
+  if (test && test.join && test.map) {
```
alanmoo: Is this effectively a browser compatibility check?

src/pages/one-page.js
```diff
@@ -3,6 +3,10 @@ import MozillaFooter from '../components/mozilla/footer.js';
 import Header from '../components/header.js';
 import SingleForm from '../components/single-form.js';
 
+// NOTE: The main page is currently about.js, due to
```
alanmoo: Where's this from?

src/lib/queryParser.js
Pomax: the `test.map` is probably not even required. The code that it wraps is for "if this was an array, turn it into a string", so just testing for "does `test.join` exist?" is probably sufficient.

src/pages/one-page.js
Pomax: from @ScottDowne

src/lib/queryParser.js
Pomax: removed test.map

src/pages/one-page.js
```diff
@@ -12,7 +16,15 @@ var onePage = React.createClass({
     return (
       <div className={className}>
         <Header alt={this.context.intl.formatMessage({id: 'donate_to_mozilla'})}></Header>
-        <SingleForm/>
+        <SingleForm
```
alanmoo: Why did this suddenly get lots o' props?

Pomax: because it was missing, and so it was just showing whatever magical defaults were baked in.

Issue #2129, opened 2018-12-17
Close out donation amount ordering test.  Implement copy test.


WillatMozFdn:
Once we're done and have a result from the donation amounts ordering test (likely Dec. 21), next test to run from Dec. 21 to Dec. 28 **(EN only)** will be copy on the left side of the page.

Current control:

**We are proudly non-profit,** non-corporate and non-compromised. Thousands of people like you help us stand up for a healthy internet for all. We rely on donations to carry out our mission to keep the Web open and free. Will you give today?

New version to test:

With a deep respect for our supporters’ privacy and a unique global reach, the non-profit Mozilla Foundation fights every day for a healthier internet, through projects like our *Privacy Not Included buyer's guide. **This work depends on contributions from people like you; will you chip in today?**

We'll need to plan to conclude the test and default all traffic onto the winner on Dec. 28th for the final EOY push.

alanmoo:
@Pomax can you queue this up (as much as is possible before we have the ordering results)?

Pomax:
Can do

Pomax:
@WillatMozFdn is it important to have a semi-colon rather than a full stop? Would **This work depends on contributions from people like you. Will you chip in today?** be acceptable?

WillatMozFdn:
semicolon not important, full stop is fine by me -- thanks @Pomax

PR #2130, opened 2018-12-17
Add A/B testing for our about sidebar text in EN locales

Pomax:
This adds a new string to the messages.properties for `en-US` only, as per https://github.com/mozilla/donate.mozilla.org/issues/2129, and adds an A/B test that does a 50/50 content swap between the original text and the new text.

### DIFF
--- a/assets/js/ab-tests.js
+++ b/assets/js/ab-tests.js
@@ -22,7 +22,6 @@ if (typeof Mozilla !== "undefined" && Mozilla.TrafficCop) {
     }
   })).init();
 
-  /*
   // This test runs on the main page (Currently about.js), for en-US only
   (new Mozilla.TrafficCop({
     id: "side-text",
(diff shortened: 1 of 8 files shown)


alanmoo:
I would assume so too. @WillatMozFdn can you confirm?

WillatMozFdn:
Yes, that would be ideal if we can.  Thanks @TheoChevalier

TheoChevalier:
Sounds good, let’s merge this sooner rather than later then :)

Pomax:
We can even land it immediately if we want, with the original text set to 100%, if we don't want to start testing the new text until the 21st. The code knows how to handle more than one active test at a time.

Pomax:
I have no idea what Travis is doing. Somehow webpack's failing even though it shouldn't because it's supposedly using webpack 3.2.0, which is the correct version.

alanmoo:
In addition to the webpack stuff, it seems like there's a handful of linting issues with the new file.

Pomax:
assuredinfantilejaguar-small

PR #2131, opened 2018-12-20
Side test ab minus iefix

alanmoo:
Since the final commit in #2130 to fix IE stuff is what broke Travis (and it seems the current "IE is running AB tests" is in production, I want to land this branch, which is #2130 without its most recent commit, plus reduction of the AB test implemented here to 0% while the other test finishes running. This way localizers can get their hands on this prior to flipping the switch next week. 

cc @Pomax

### DIFF
--- a/assets/js/ab-tests.js
+++ b/assets/js/ab-tests.js
@@ -22,7 +22,6 @@ if (typeof Mozilla !== "undefined" && Mozilla.TrafficCop) {
     }
   })).init();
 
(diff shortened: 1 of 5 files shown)


TheoChevalier:
👍 I can also land the string separately in case there’s further issues

alanmoo:
Looks like this is passing, so I'm going to give it one more review on the test server and land it.

PR #2162, opened 2019-09-18
Fix #2161 - don't (try to) pass window.grecaptcha.execute to expectRecaptcha which is going to wait for grecaptcha to become awailable.

### LINKED ISSUE #2161: donating doesn't work (window.grecaptcha undefined)
Got a report from a (want to be) donor that the donate by credit card button will just keep spinning, so donating can't proceed. He had attached a screenshot showing the trace for window.grecaptcha undefined.

Took a look at the code, and I see an obvious problem in executeRecaptcha.

mkmelin:
At the time window.grecaptcha could still be unavailable. So, only access it after we know it's there.

I think this should work. But I don't get a captcha when testing locally with or without this patch... so unable to verify.

### DIFF
--- a/src/mixins/stripe.js
+++ b/src/mixins/stripe.js
@@ -88,9 +88,10 @@ var StripeMixin = {
       return;
     }
     if (!window.grecaptcha) {
-      return setTimeout(() => {
+      setTimeout(() => {
         this.expectRecaptcha(callback);
       }, 100);
+      return;
     }
 
     window.grecaptcha.ready(callback);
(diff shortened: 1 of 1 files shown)


TheoChevalier:
CC @alanmoo do you know who could review this?

src/mixins/stripe.js
```diff
@@ -88,9 +88,10 @@ var StripeMixin = {
       return;
     }
     if (!window.grecaptcha) {
-      return setTimeout(() => {
+      setTimeout(() => {
         this.expectRecaptcha(callback);
       }, 100);
+      return;
```
cadecairos: Why this change? It doesn't appear to do anything

mkmelin: Strictly not necessary, it doesn't change the behaviour. But it's not good practice to have methods that sometimes return something (well undefined, so not much) and sometimes not, as some linters will tell you...

